PR #152, opened 2019-01-15
Update js to skip servers without monitoring data

gonzalez:
### Description

Skips any operational vm that is not sending cpu or memory monitoring data back to the platform.

### Issues Resolved

https://github.com/rightscale/policy_templates/issues/151

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/cost/downsize_instance/downsize-instance.pt
+++ b/cost/downsize_instance/downsize-instance.pt
@@ -382,10 +382,10 @@ policy "policy_rightsize" do
     detail_template <<-EOS
 # {{ len data }} Instances can be downsized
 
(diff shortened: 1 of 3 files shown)


gonzalez:
> what is the reason for the account or account id in the incident. Can we remove it instead of avoiding the | issue?

You can run the policy across accounts.

cost/downsize_instance/downsize-instance.pt
```diff
@@ -1,8 +1,8 @@
-name "Downsize Instances based on CPU and Memory"
+name "Downsize Instances based on CPU and Memory Update"
```
cdwilhelm: I don't understand how `Update` applies in this name.

gonzalez: Carry over from testing.  Fixed.  Thanks

PR #167, opened 2019-01-29
Azure Reserved Instances Policy

ryanoleary:
### Description

This adds a new policy for reporting on underutilized Azure Reserved Instances

### Issues Resolved

none

### Contribution Check List

- [ ] All tests pass.
- [x ] New functionality has been documented in the README if applicable
- [x ] New functionality has been documented in CHANGELOG.MD

### DIFF
--- a/cost/azure/reserved_instances/utilization/azure_reserved_instance_utilization.pt
+++ b/cost/azure/reserved_instances/utilization/azure_reserved_instance_utilization.pt
@@ -1,0 +1,157 @@
+name "Azure Reserved Instances Utilization"
+rs_pt_ver 20180301
+type "policy"
+short_description "A policy that sends email notifications when utilization falls below a threshold. See the [README](https://github.com/rightscale/policy_templates/tree/master/cost/azure/reserved_instances/utilization) and [docs.rightscale.com/policies](http://docs.rightscale.com/policies/) to learn more."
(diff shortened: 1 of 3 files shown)


ryanoleary:
Note that testing was performed against an Azure EA that had RIs.

cost/azure/reserved_instances/utilization/azure_reserved_instance_utilization.pt
```diff
@@ -0,0 +1,157 @@
...
+script "azure_ri_request", type: "javascript" do
+  parameters "enrollment"
+  result "request"
+  code <<-EOS
+
+    var date = new Date();
+    date.setDate(date.getDate() - 3);
+    var enddate = date.toISOString().slice(0,10);
+    date.setDate(date.getDate() - 7);
+    var startdate = date.toISOString().slice(0,10);
```
douglaswth: There should probably be a comment explaining this date math.

ryanoleary:
@cdwilhelm  i can't reply in your comment thread, but do you mean the Azure API access? If so, I don't think we need to comment on that because that key always has this access as far as I know. That key provides a broad range of access to the EA APIs.

cdwilhelm:
@ryanoleary ok, sounds fine.  thanks.

PR #186, opened 2019-02-26
Allow for chart dimension to be selectable

ryanoleary:
Only show top 10 categories (+Other) in chart

### Description

Allows for the report dimension to be selected from any of the "standard" dimensions
Changes the chart to only show the top 10 values with all other data grouped into "Other" (like the UI does)

### Issues Resolved

N/A

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/cost/scheduled_reports/scheduled_report.pt
+++ b/cost/scheduled_reports/scheduled_report.pt
@@ -48,8 +48,8 @@ end
 parameter "param_graph_dimension" do
   type "string"
   label "Graph Dimension"
-  allowed_values "category", "service", "instance_type", "region", "resource_type", "vendor", "vendor_account"
-  default "category"
+  allowed_values "Category","Service","Instance Type","Region","Resource Type","Cloud Vendor","Cloud Account"
+  default "Category"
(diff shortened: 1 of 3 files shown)


cost/scheduled_reports/scheduled_report.pt
```diff
@@ -535,7 +535,7 @@ script "js_generate_report", type: "javascript" do
     }
 
     var chartData = "chd=a:" + chartDataArray.join('|')
-    var encodedCategories = encodeURIComponent(categories.join('|'))//.replace(/%7C/gi,'|');
+    var encodedCategories = encodeURIComponent(categories.join('|')).replace(/[(]/gi,'%28').replace(/[)]/gi,'%29');
```
ryanoleary: Question - with this change not only have you encoded (), but you've undone something that looks like it unencoded | . was that intentional? and will it be ok?

srpomeroy: If you look carefully, the unencoding of `|` was actually commented out. I missed that when cleaning up my debug code during the initial PR of the chart enhancement.

PR #194, opened 2019-03-11
Azure: Tag Resources with Resource Group Name

dfrankel33:
### Description

Initial release of:  **Azure: Tag Resources with Resource Group Name**

### Issues Resolved
N/A

### Contribution Check List

- [x] All tests pass.


compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,314 @@
...
+pagination "azure_pagination" do
+  get_page_marker do
+    body_path "nextLink"
+  end
+  set_page_marker do
+    uri true
+  end
+end
+
+datasource "azure_resources" do
```
srpomeroy: Datasource names should begin with `ds_`

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,314 @@
...
+    collect jmes_path(response, "value[*]") do
+      field "id", jmes_path(col_item,"id")
+      field "region", jmes_path(col_item,"location")
+      field "name", jmes_path(col_item,"name")
+      field "tags", jmes_path(col_item,"tags")
+    end
+  end
+end
+
+datasource "filtered_resources" do
```
srpomeroy: Datasource names should begin with `ds_`

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,314 @@
...
+            })
+          }
+        }
+      }
+    })
+
+EOS
+end
+
+policy "azure_resource_policy" do
```
srpomeroy: Policy names should begin with `policy_`

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,314 @@
...
+            resource_type: resource_type,
+            tag: 'none'
+          })
+        } else {
+          if (_.has(resource_tags, param_tag_key)) {
+            var rg_tag_value = resource_tags[param_tag_key] ;
+            console.log('TAG KEY FOUND')
+            console.log('Tag Value: '+rg_tag_value)
+
+            if (resource_group_name == rg_tag_value) {
```
srpomeroy: This if/else statement seems redundant as the check is happening in the policy block.

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,314 @@
...
+
+escalation "tag_resources" do
+  run "tag_resources", data, $param_azure_tenant_id, $param_azure_sub, $param_tag_key
+end
+
+define tag_resources($data, $param_azure_tenant_id, $param_azure_sub, $param_tag_key) return $all_responses do
+
+  call get_access_token($param_azure_tenant_id) retrieve $access_token
+
+  $providers_response = http_request(
```
srpomeroy: If I remember correctly, this is a very expensive call. Curious if you have any benchmark data on how long this call takes?

srpomeroy:
Please also update the root README.md file

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
dfrankel33: yep. was remnant from debugging. good catch.

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
dfrankel33: i do not

srpomeroy: How about how long the policy takes to run from start to end in an example account?

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,300 @@
...
+        }
+      }
+    })
+
+EOS
+end
+
+policy "policy_azure_resource" do
+  validate_each $ds_filtered_resources do
+    summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{ len data }} Azure Resources out of compliance"
```
srpomeroy: If its not specifically targeting an RS account, suggest removing this information from the summary_template. Could use the subscription id param instead.

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,300 @@
...
+    detail_template <<-EOS
+# Azure Resources Incorrectly Tagged
+| Resource Name | Resource Group | Azure Region | Resource ID | Tag Value | 
+| ------------- | -------------- | ------------ | ----------- | --------- |
+{{ range data -}}
+| {{.name}} | {{.resource_group}} | {{.region}} | {{.id}} | {{.tag}}
+{{ end -}}
+EOS
```
srpomeroy: If not focusing on an RS account, recommend adding a footer to the detail template with information about which RS account its running in. Example: https://github.com/rightscale/policy_templates/blob/master/cost/scheduled_reports/scheduled_report.pt#L590-L592

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,303 @@
...
+  category "Azure"
+end
+
+parameter "param_azure_sub" do
+  type "string"
+  label "Azure Subscription ID"
+  category "Azure"
+end
+
+parameter "param_tag_key" do
```
psschroeter: Is this tag_key parameter always the same? If not, when does it differ. If so, shouldn't it be a constant embedded in code or at least have a default value?
Also FYI parameters do support a "description" field -- The api certainly returns it I think the UI will display it inline rather than only having it embedded in the source.

dfrankel33: it's going to be customer-specific, so it should be same within a single org

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,303 @@
...
+          query_strings: {
+            "api-version": $api_version
+          },
+          headers: {
+            "cache-control": "no-cache",
+            "content-type": "application/json",
+            "authorization": "Bearer " + $access_token
+          },
+          #body: $new_object
+          body: {"tags": $new_tags}
```
avinashbhashyam-rs: I am guessing you have to reset tags here, can't just add an additional tag. My 2 Cents you are already fetching the resource tags in the javascript data source. Add an additional field to the results to store existing tags and reuse it here while adding a new tag instead of an additional API call to fetch tags.

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
```diff
@@ -0,0 +1,303 @@
...
+  $providers_array = $providers_response["body"]["value"]
+  $$log = []
+  $all_responses = []
+  foreach $item in $data do
+    $$log << "ID: "+$item["id"]
+    $namespace = $item["namespace"]
+    $$log << "Namespace: "+$namespace
+    if $namespace =~ "Classic"
+      # Skip
+      $$log << "Skipping due to Azure Classic"
```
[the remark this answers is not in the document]
dfrankel33: nope.  i'll remove it

compliance/tags/azure_rg_tags/azure_resource_group_tags.pt
dfrankel33: good call! fixed.

PR #222, opened 2019-05-01
adding in codeowners

rshade:
### Description

This adds in CODEOWNERS into the repo to automate reviews. 

### Issues Resolved

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,11 @@
 # This is a comment.
+
 # Each line is a file pattern followed by one or more owners.
-*.pt @rightscale/professional-services @rightscale/phoenix
+
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+*.pt @rightscale/professional-services @rightscale/phoenix
+* @rightscale/professional-services
```
srpomeroy: I think the rule with * needs to be first as the last rule to match always wins.

rshade: I would think that means professional-services gets notified on everything, and *.pt notifies both teams.

srpomeroy: It verifies each rule separately. It doesn't stop as soon as a rule matches, but rather at EOF. So the last match wins. So the current way you have it means the PS time will always be notified and the previous rules essentially don't matter.

PR #242, opened 2019-05-24
Part One Splitting Downsize policy into RL utilization and Multi cloud downsize

cdwilhelm:
### Description

Part one of Splitting the RL downsize and RL Utilization into two policies.  One for Multi Cloud downsize and one for RL Utilization.  This part is for the downsize.  The RL Utilization will come shortly.

### Contribution Check List

- [x] All tests pass.


srpomeroy:
@cdwilhelm as discussed, we need to avoid resizing Azure instances that are part of an availability set. Not sure if we should report on it, or just ignore it all together like this: https://github.com/rightscale/policy_templates/blob/2146453ff59f8e97084444ac84f2414ba3ec9897/cost/azure/superseded_instance_types/azure_superseded_instance_types.pt#L106

srpomeroy:
> @cdwilhelm as discussed, we need to avoid resizing Azure instances that are part of an availability set. Not sure if we should report on it, or just ignore it all together like this:
> 
> https://github.com/rightscale/policy_templates/blob/2146453ff59f8e97084444ac84f2414ba3ec9897/cost/azure/superseded_instance_types/azure_superseded_instance_types.pt#L106

That example is using the Azure API. In CM it appears in cloud-specific attributes: $instance.cloud_specific_attributes.availability_set

PR #405, opened 2019-11-25
POL-86 FSM Suspicious Users

dfrankel33:
### Description

initial release of Suspicious Users policy

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/saas/fsm/suspicious_users/README.md
+++ b/saas/fsm/suspicious_users/README.md
@@ -2,7 +2,7 @@
 
 ## What it does
 
-This policy will raise an alert when Flexera SaaS Manager identifies suspicious users logging into SaaS applications.
+This policy will create an incident when Flexera SaaS Manager identifies suspicious users logging into SaaS applications.
 
 ## Functional Description
 
--- a/saas/fsm/suspicious_users/README.md
+++ b/saas/fsm/suspicious_users/README.md
@@ -1,0 +1,28 @@
+# SaaS Manager - Suspicious Users
+
+## What it does
+
+This policy will raise an alert when Flexera SaaS Manager identifies suspicious users logging into SaaS applications.
+
+## Functional Description
+
+This policy integrates with the Flexera SaaS Manager API to retrieve suspicious activities. Therefore the following are prerequisites for this policy to execute:
+
+- Flexera SaaS Manager implementation with HR roster connected
+- Retrieve a Flexera IAM refresh token for authentication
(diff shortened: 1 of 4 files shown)


saas/fsm/suspicious_users/fsm-suspicious_users.pt
```diff
@@ -0,0 +1,149 @@
+name "SaaS Manager - Suspicious Users"
+rs_pt_ver 20180301
+type "policy"
+short_description "This policy will raise an alert when Flexera SaaS Manager identifies suspicious users logging into SaaS applications. See the [README](https://github.com/flexera/policy_templates/tree/master/saas/fsm/suspicious_users/) and [docs.rightscale.com/policies](http://docs.rightscale.com/policies/) to learn more."
```
cdwilhelm: product doesn't want to use alert.  instead change **raise an alert** to **create an incident**.

saas/fsm/suspicious_users/README.md
```diff
@@ -0,0 +1,28 @@
+# SaaS Manager - Suspicious Users
+
+## What it does
+
+This policy will raise an alert when Flexera SaaS Manager identifies suspicious users logging into SaaS applications.
```
cdwilhelm: product doesn't want to use alert.  instead change **raise an alert** to **create an incident**.

saas/fsm/suspicious_users/fsm-suspicious_users.pt
```diff
@@ -0,0 +1,149 @@
...
+           field "unauthorized", val(col_item, "unauthorized")
+        end
+    end
+end
+
+datasource "ds_users_cleanup" do
+  run_script $js_users_cleanup, $ds_suspicious_users
+end
+
+script "js_users_cleanup", type: "javascript" do
```
cdwilhelm: apply some sorting, maybe by Email address.

dfrankel33: this isn't the point of the policy.  there should never be any users returned from this call. if there are, it should raise an incident

cdwilhelm: If there is a list of users shouldn't the list be sorted in some way?

dfrankel33: i don't see why.  there is no way to identify priority, and every user in this list should be included in the incident

PR #420, opened 2019-12-11
POL-163 read info field

cdwilhelm:
### Description

* add policyparser to parse fields from policy templates.
* update the Rakefile to create the active policy list to read the info field hash.
* update the Dangerfile to check the info field and it's keys.  version and provider are required.

### Issues Resolved

### Contribution Check List

- [ ] All tests pass.


Dangerfile
```diff
@@ -124,6 +124,30 @@ has_app_changes.each do |file|
   end
 end
 
+# check for info field required fields
+has_app_changes.each do |file|
+  # get info field data
+  info = {}
+  content = File.read(file)
+  info_body = content.scan(/info\((.*?)\)/m)
```
douglaswth: Although, this may work fine for the data that you are currently being used in the info hashes, this method of getting the info hash out of policy templates seems like it could be quite error prone if someone tries to put arbitrary strings into some value in the hash that could contain a `)`. A more robust approach would be to create a small Ruby DSL that mimics the Policy Template language but only actually implements the `info` method and uses it to collect the info hash.

Rakefile
```diff
@@ -12,17 +12,31 @@ task :generate_policy_list do
...
     readme = ::File.join(file.split('/')[0...-1].join('/'),'README.md')
-    @version = nil
 
     if !file.match(/test_code/)
       f = File.open(file, "r:bom|utf-8")
+
+      # get info field data
+      info = {}
+      content = File.read(file)
+      info_body = content.scan(/info\((.*?)\)/m)
```
douglaswth: This has the same issues as I previously mentioned above.

Rakefile
```diff
@@ -49,12 +63,15 @@ task :generate_policy_list do
...
+        "version": version,
         "change_log": change_log,
         "description": @description,
         "category": @category,
         "severity": @severity,
-        "readme": readme
+        "readme": readme,
+        "provider":provider,
+        "service":service,
+        "policy_set":policy_set
```
douglaswth: This is more of a style comment than anything else, but usually there is a space after the `:` and having a trailing `,` here would prevent the extra diff line above from happening in the future.

Dangerfile
cdwilhelm: @douglaswth This is a great suggestion.  I tested this exact scenario and it caused the build to fail, because it fails to evaluate the string.  With that, I'm not concerned about this since the build fails and would force the user to correct before we'd approve..

The DSL is something I'd like to consider though.  Do you have any suggestion where to start?  An example perhaps?

douglaswth: Here's a DSL that does the job as long as the `info` hash is defined before any RCL `define` declarations (I imagine everyone on the content team will be putting the metadata before any declarations).

```ruby
class DSL
  DEFINITION_REGEX = /^[[:blank:]]*define[[:blank:]]*([\w_\.]+)[[:blank:]]*\([@$\w _,]*\)[[:blank:]]*.*do/.freeze

  attr_reader :parsed_info

  def parse(file)
    source = File.read(file)

    # we assume that all of the metadata is before any declarations
    # especially RCL defines which are not parsable as Ruby
    source = source.split(DEFINITION_REGEX, 2).first

    instance_eval(source, file)
  end

  def info(**info)
    @parsed_info = info
  end

  def method_missing(symbol, *args, &block)
    # do nothing
  end
end

dsl = DSL.new
dsl.parse('path/to/policy_template.pt')

p dsl.parsed_info
```

cdwilhelm: Thanks @douglaswth.  this is awesome.

cdwilhelm: @douglaswth  I updated the PR to use the DSL.  I created a lib file to use in the Dangerfile and Rakefile and added other fields to parse.  so much nicer than using regex.  
Can you give another review?

Rakefile
```diff
@@ -1,6 +1,7 @@
 require 'rubygems'
 require 'json'
 require 'fileutils'
+require './tools/lib/policy_parser'
```
douglaswth: Same as above regarding `require_relative`.

cdwilhelm:
@douglaswth  I made the changes for the require_relative.  can you give one more review/approve?

PR #470, opened 2020-03-03
Phx 2039 refactor tool run

psschroeter:
Notes:
This list was generated by doing:
```
find . -name '*.pt' -exec ruby ./tools/bin/convert_table {} \;
```
This is a do-not-merge, just putting this up to make sure we're not missing anything from the data table. I'd advise re-running the tool on a more small scale as needed to get started when the time is right for it.

The tool is just a rough first pass to help avoid typos and mistakes and therefore hopefully speed up testing. Various issues notes in comments.

### DIFF
--- a/compliance/azure/ahub_manual/azure_ahub_utilization_with_manual_entry.pt
+++ b/compliance/azure/ahub_manual/azure_ahub_utilization_with_manual_entry.pt
@@ -292,13 +292,44 @@ policy 'azure_ahub_policy' do
     summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{data.summary_template}}"
     detail_template <<-EOS
 # Azure AHUB Utilization with Manual Entry
-| Subscription Name | Location | Resource Group | Instance Name | Instance Size | Image Publisher | Image Offer | Image Sku | Image Version |  Tags | Number Cores |
-| ------------- | -------------- | -------- | ------------- | --------------- | ----------- | --------- | ------------- |  ---- | ------------ |
-{{ range data.result -}}
-|{{.subscriptionName}} | {{.location}} | {{.resourceGroup}} | {{.name}} | {{.vmSize}} | {{.imagePublisher}} | {{.imageOffer}} | {{.imageSku}} | {{.imageVersion}} | {{.tags}} | {{.numberofcores}} |
-{{ end -}}
 ##### Note: {{data.note}}
   EOS
+    export 'result' do
+      field "subscriptionName" do
+        label "Subscription Name"
+      end
+      field "location" do
+        label "Location"
+      end
+      field "resourceGroup" do
+        label "Resource Group"
+      end
+      field "name" do
+        label "Instance Name"
+      end
+      field "vmSize" do
+        label "Instance Size"
+      end
+      field "imagePublisher" do
+        label "Image Publisher"
+      end
+      field "imageOffer" do
+        label "Image Offer"
+      end
+      field "imageSku" do
+        label "Image Sku"
+      end
+      field "imageVersion" do
+        label "Image Version"
+      end
+      field "tags" do
+        label "Tags"
+      end
+      field "numberofcores" do
+        label "Number Cores"
+      end
+    end
+
   escalate $email
   check eq(size(val(data, "result")), 0)
   end
--- a/compliance/tags/tag_checker/tag_checker.pt
+++ b/compliance/tags/tag_checker/tag_checker.pt
@@ -291,12 +291,39 @@ policy "untagged_resources_policy" do
     detail_template <<-EOS
 There are {{ len data }} resource(s) missing tags: {{ parameters.param_tag_keys }}
 
-| Cloud Name | Cloud Type | Name | ID | Type | State | Href | Missing Tags | Invalid Tags |
-| ---------- | ---------- | ---- | --- | ---- | ----- | ---- | ------------ | ------------ |
-{{ range data -}}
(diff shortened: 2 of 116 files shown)


compliance/azure/ahub_manual/azure_ahub_utilization_with_manual_entry.pt
```diff
@@ -292,13 +292,44 @@ policy 'azure_ahub_policy' do
...
     detail_template <<-EOS
 # Azure AHUB Utilization with Manual Entry
-| Subscription Name | Location | Resource Group | Instance Name | Instance Size | Image Publisher | Image Offer | Image Sku | Image Version |  Tags | Number Cores |
-| ------------- | -------------- | -------- | ------------- | --------------- | ----------- | --------- | ------------- |  ---- | ------------ |
-{{ range data.result -}}
-|{{.subscriptionName}} | {{.location}} | {{.resourceGroup}} | {{.name}} | {{.vmSize}} | {{.imagePublisher}} | {{.imageOffer}} | {{.imageSku}} | {{.imageVersion}} | {{.tags}} | {{.numberofcores}} |
-{{ end -}}
 ##### Note: {{data.note}}
   EOS
+    export jmes_path(data, 'result') do
```
psschroeter: Note: We're changing this from:
```
export "result" do
```
to
```
export jmes_path(data, "result") do
```
After some discussion on phoenix team, for a couple reasons:
* Support for jq() instead of jmes_path() was added and needed a way to switch
* Future flexibility

compliance/tags/tag_checker/tag_checker.pt
```diff
@@ -291,12 +291,41 @@ policy "untagged_resources_policy" do
...
+      field "type" do
+        label "Type"
+      end
+      field "state" do
+        label "State"
+      end
+      field "href" do
+        label "Href"
+      end
+      field "range__i___e_____missing_tags" do
```
psschroeter: Auto-conversion didn't work so well for this case -- will have to hand fix these type of things.
I think it should just be:
```
field "missing_tags" do
  label "Missing Tags"
end
```

These is untyped json data -- it can be boolean, float/int, string, an array, or a map. The UI will know how to display all those values except map is up in the air, I don't know what the UI might do with that. The UI should know to comma join arrays of simple types...

cost/aws/burstable_instance_cloudwatch_credit_utilization/aws_burstable_instance_cloudwatch_credit_utilization.pt
```diff
@@ -566,14 +566,55 @@ policy "pol_utilization" do
...
+      field "privateDnsName" do
+        label "Private DNS Name"
+      end
+      field "cpu_maximum" do
+        label "CPU Maximum %"
+      end
+      field "cpu_average" do
+        label "CPU Average %"
+      end
+      field "if_eq__cloudwatch_cpu_credit_balance_min__10000_" do
```
psschroeter: Will just have to move this logic to the javascript I think....

cost/superseded_instance/superseded_instance.pt
```diff
@@ -413,15 +413,42 @@ policy "pol_aws_superseded_instance" do
...
-{{ end -}}
-
 ___
 ###### Policy Applied in Account: {{ rs_project_name }} (Account ID: {{ rs_project_id }}) within Org: {{ rs_org_name }} (Org ID: {{ rs_org_id }})
 EOS
+    export do
+      field "if__warn" do
+        label "Resource UID"
+        path jmes_path(item, "if .warn")
+        format "printf(%s**_ end -}} .resource_id }} if .warn }}_** end -}})"
```
psschroeter: Ideally we might do something like
```
field "resource_uid" do
  label "Resource UID"
  format switch(val(item, "warn"), "bold", "")
end
```

But the "format" field currently doesn't accept functions -- might change this on the backend to support this, though it currently doesn't work like this.

cost/aws/savings_plan/recommendations/aws_savings_plan_recommendations.pt
```diff
@@ -208,19 +208,57 @@ policy "aws_sp_recommendations" do
...
-{{ range data -}}
-| {{.estimatedMonthlySavingsAmount}} ({{.currencyCode}}) | {{.estimatedSavingsPercentage}}% | {{.estimatedSavingsPlanCost}} ({{.currencyCode}}) | {{.hourlyCommitmentToPurchase}} ({{.currencyCode}}) | {{.upfrontCost}} ({{.currencyCode}}) | {{.instanceFamily}} | {{.region}} | {{.currentAverageHourlyOnDemandSpend}} ({{.currencyCode}}) | {{.currentMaximumHourlyOnDemandSpend}} ({{.currencyCode}}) | {{.currentMinimumHourlyOnDemandSpend}} ({{.currencyCode}}) | {{.offeringId}}
-{{ end -}}
-
 ###### Policy Applied in Account: {{ rs_project_name }} (Account ID: {{ rs_project_id }}) within Org: {{ rs_org_name }} (Org ID: {{ rs_org_id }})
 EOS
+    export do
+      field "estimatedMonthlySavingsAmount" do
+        label "Estimated Monthly Savings"
+        format "printf(%s (.currencyCode}}))"
```
psschroeter: Hmm not sure about this -- maybe just keep it simple and make currency code a separate field.

compliance/azure/ahub_manual/azure_ahub_utilization_with_manual_entry.pt
psschroeter: Err, after talking with Curt, I going to have it be the older format (text strings)... will update this.

PR #561, opened 2020-07-21
POL-339-Updated Readme

chandravurubindi:
POL-339-Updated Readme

### Description

Updated credential details in README file

### Issues Resolved

Updated README file with OAuth2 type as Credential Service

### Contribution Check List

- [ ] All tests pass.


chandravurubindi:
> Do you think this is enough detail for a user that read the credential docs to create a new credential for the storage account?
>  I think the details what was mentioned in the credentials docs for the storage account is sufficient and updated the readme file by specifying API Usage section in the credentials docs for user reference. 
> if we would like to add more details for the storage account, then it might be better some updation for Azure API Usage section in the credentials docs as follows:
In Flexera Governance, navigate to the Credentials page, press the New Credential button, and select the oauth2 credential type.
Providing a screenshot for oauth2 type.

> Also when viewing the rendered page the url doesn't render correctly. I suggest removing all the < and >. If a user used your example it wouldn't work.
> Handled this issue by adding ' and '.

cdwilhelm:
> > Do you think this is enough detail for a user that read the credential docs to create a new credential for the storage account?
> > I think the details what was mentioned in the credentials docs for the storage account is sufficient and updated the readme file by specifying API Usage section in the credentials docs for user reference.
> > if we would like to add more details for the storage account, then it might be better some updation for Azure API Usage section in the credentials docs as follows:
> > In Flexera Governance, navigate to the Credentials page, press the New Credential button, and select the oauth2 credential type.
> > Providing a screenshot for oauth2 type.
> 
> > Also when viewing the rendered page the url doesn't render correctly. I suggest removing all the < and >. If a user used your example it wouldn't work.
> > Handled this issue by adding ' and '.

@chandravurubindi   I'm happy with the changes.  thank you.

PR #569, opened 2020-07-29
POL-344-Budget-Alerts-w/-Billing-Center-Unallocated

KVivek03:
### Description

POL-344-Budget-Alerts-w/-Billing-Center-Unallocated

### Issues Resolved

Added filter condition to get value only for Parent BC.

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/cost/budget_alerts/CHANGELOG.md
+++ b/cost/budget_alerts/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.13
 
-- Added the filteration logic for getting only the Parent BC details
+- Added the ability to provide either Billing center name or Id.
 
 ## v1.12
 
--- a/cost/budget_alerts/CHANGELOG.md
+++ b/cost/budget_alerts/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.13
+
+- Added the filteration logic for getting only the Parent BC details
+
 ## v1.12
 
 - Fixed missing budget alert incident graph
--- a/cost/budget_alerts/budget_alert.pt
+++ b/cost/budget_alerts/budget_alert.pt
@@ -8,7 +8,7 @@ category "Cost"
 tenancy "single"
 default_frequency "daily"
 info(
-  version: "1.12",
+  version: "1.13",
   provider: "Flexera Optima",
   service: "",
   policy_set: ""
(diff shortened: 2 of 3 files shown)


cost/budget_alerts/budget_alert.pt
```diff
@@ -257,10 +257,12 @@ end
 
 script "js_filtered_billing_centers", type: "javascript" do
   parameters "billing_centers", "param_bc_name"
-  result "filtered_billing_centers"
+  result "filtered_billing_centers_new"
   code <<-EOS
   var filtered_billing_centers =
     _.reject(billing_centers, function(bc){ return (param_bc_name !== bc.name) });
+  var filtered_billing_centers_new =
+    _.reject(filtered_billing_centers, function(bc){ return (bc.parent_id != null) })
```
cdwilhelm, line 265: I'm not sure if this is what we want.  Can you explain the logic?  It would be helpful to include some comments about what the new line is doing for anybody reading the code.  Also do you get the same results of the billing centers that are not Unallocated?

cost/budget_alerts/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.13
+
+- Added the filteration logic for getting only the Parent BC details
```
cdwilhelm, line 5: I thought the goal was to get the parent BC for Unallocated.  Also the README should probably be updated to include any changes in the filtering it's doing.

cdwilhelm: Also you if you haven't already you should compare the old report with your new report and make sure it matches your expectations.  Also compare to Optima.

cost/budget_alerts/budget_alert.pt
KVivek03: Hi @cdwilhelm , This new line is filtering out the child BC and we are left with only the Parent BC. As in case of unallocated BC, the child Unallocated BC will be filtered out. I will add the comment in the policy.

cost/budget_alerts/CHANGELOG.md
cdwilhelm: What about other Billing Centers that are not Unallocated.  What is the new behavior going to do with them?

KVivek03: Hi @cdwilhelm , The issue may appear if  suppose there is a parent BC with name `Test` and multiple child BC with similar name, So i have included the filter condition to get value of Parent BC only. Please let me know if i should make the change only for `Unallocated` only. I have compared the result using an older policy which was in the Rightscale portal and the latest one, and i have checked the same in the Optima also. PFB snapshot of the same:
This is with the older policy:
image

This is of the modified policy:
image

and this is the value from optima:
image
Thanks

KVivek03: For other BC also, it will fetch the values from Parent BC only, and there should not be any data mismatch, like the unallocated BC. PFB snap of the values of `Test` BC :
from older policy:
image

from the modified policy:
image

in the optima:
image
Thanks

KVivek03: Here in the above case, since there was no child BC with name `Test`, there is similar data in both the older and the newer policy. Thanks. Should i change the filter condition for BC `Unallocated` only? If there are other multiple BC with similar name, in that case data mismatch can be there. Thanks
cc @chandravurubindi

PR #580, opened 2020-08-10
Pol 356 aws schedule instances action not working

KVivek03:
### Description

modified the policy to handle start or stop time values without minute value.

### Contribution Check List

- [ ] All tests pass.


cost/aws/schedule_instance/aws_schedule_instance.pt
```diff
@@ -84,6 +84,7 @@ datasource "ds_get_all_instances" do
     query "Filter.1.Name", "tag-key"
     query "Filter.1.Value.1", "schedule"
     query "Version", "2016-11-15"
+    ignore_status [403]
```
cdwilhelm: We should removed this because it will result in no instances and no indication of an error to inform the user something is wrong.

KVivek03: made the change

PR #619, opened 2020-10-21
Fix AWS Regions with SCP for FinOps Policies

douglaswth:
### Description

- Update the AWS Unused Volumes, AWS Old Snapshots, and AWS Unused IP
  Addresses policy templates to use the actual calls to determine if
  each region is available via SCP since we have discovered that using
  `DescribeRegions` in each particular region does not work reliably
  (the `DescribeRegions` call could be enabled for all regions while
  other calls are not).
- Add the Code Spell Checker extension to the suggested VS Code
  extensions and fix up some misspellings it found.
- Removed confusing ignored `host "ec2.amazonaws.com"` lines from all of
  the `ds_regions` datasources.

### Issues Resolved

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/compliance/aws/ecs_unused/aws_unused_ecs_clusters.pt
+++ b/compliance/aws/ecs_unused/aws_unused_ecs_clusters.pt
@@ -75,7 +75,6 @@ datasource "ds_regions" do
     request do
       auth $auth_aws
       verb "GET"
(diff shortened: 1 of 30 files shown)


cdwilhelm (approved):
LGTM

PR #642, opened 2020-12-01
AWS Service Control Policy - Pagination Fix

srpomeroy:
### Description

Fixes pagination to add the `NextToken` to the body instead of query params

### Issues Resolved

N/A

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/compliance/aws/scp_audit/aws_scp_audit.pt
+++ b/compliance/aws/scp_audit/aws_scp_audit.pt
@@ -5,9 +5,9 @@ short_description "This policy checks to see if the provided service control pol
 category "Compliance"
 severity "medium"
 info(
-  version: "2.1",
+  version: "2.2",
   provider:"AWS",
-  service: "Org", 
+  service: "Org",
   policy_set: ""
 )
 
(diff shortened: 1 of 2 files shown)


compliance/aws/scp_audit/aws_scp_audit.pt
```diff
@@ -48,7 +48,7 @@ pagination "pagination_aws_json" do
     body_path jmes_path(response, "NextToken")
   end
   set_page_marker do
-    query "NextToken"
+    body_field "NextToken"
```
douglaswth, line 51: Looks good! I just noticed below that there is the following in the `ds_org_accounts` datasource:

```ruby
body_field "empty", "empy"
```

That's a little hacky, so while you are here, I think you could change it to an empty JSON object with:

```ruby
body "{}"
```

srpomeroy: You learn something new every day. Thanks @douglaswth!

douglaswth: 🤣 I actually implemented it two years ago so I was surprised to see that there were workarounds for the lack of it, but then I checked the docs and apparently it never made it in there, so now I'm going to add it there as well.

douglaswth (approved):
LGTM

PR #647, opened 2020-12-15
POL-410-Use-idle-IP-costs-for-displaying-total-savings-in-AWS-unused-IP-address-policy

KVivek03:
POL-410-Use-idle-IP-costs-for-displaying-total-savings-in-AWS-unused-IP-address-policy

Modified policy to only use Idle costs from Optima for calculating the total estimated monthly savings

### Contribution Check List

- [ ] All tests pass.


douglaswth (changes requested):
With this change it would also make sense to put the per unattached IP address estimated savings into the `export` by dividing the total estimated savings by the number of unattached IP addresses.

KVivek03:
Hi @douglaswth ,
I am not sure if the splitting would be showing the per-resource cost correctly, as for the testing purpose when I applied the policy for this account - 123812461563, We are seeing cost for only one IP from Optima
image
but there are in total 11 unused IPs. So by the suggested approach we would be showing one IP cost splitted among 11 of them. Please let me know your thoughts.

douglaswth:
That does not seem like a very useful approach. We should not just be passing whatever costs we're seeing in Optima as if they are the total estimate, but rather we should use the data in Optima to determine what the cost per month of a single unused IP address would be and then you can multiply that by the count of unused IP addresses that the policy has discovered to get the total estimated savings and you can report the individual estimated savings easily as well.

KVivek03:
Hi @douglaswth , If we are checking for monthly cost for the IPs and selecting any one of them and then calculating and dividing among the discovered unused IP by policy, It can lead to showing incorrect results as the **IPs are charged hourly. Moreover there can be few IPs unused from 15 or 20 or even 30 days.** So calculating on the basis of one and showing as savings would not be giving the correct estimated savings information. The general approach which we are using is to find for 1 day and show it for a month after multiplying with 30 which is also mentioned in README. Since Optima does not have the AWS IPs id, we show the sum total of the entire received cost data from Optima.
Please let us know your thoughts.

douglaswth:
If you cannot figure out what the estimated savings for a single unused IP address is, I don't see how the current methodology could possibly give an accurate total estimate. An estimate is a future looking calculation based on the current run rate, not something based on what has already happened previously.

If you can find the costs for 1 day, you should be able to figure out from that what the hourly rate is for a single unused IP address from that and use it to actually calculate out the estimated savings over a month that would be realized if each of the unused IP addresses detected were to be removed. If that is not actually possible currently with the data that you are getting from Optima, then I think we should consider figuring out a different solution as I don't believe the current solution is going to be accurate otherwise.

KVivek03:
Hi @douglaswth , I had a discussion with @nshivappa and we thought since the cost seems to be fixed for Unused IPs which is $0.005 per hour so should we show the estimated savings as per this data for the remaining days. For example if user ran the policy , so our policy will consider remaining days of the current month(13 days if ran today)*24(hour)*0.005 to compute the estimated savings per-resource for this month and then we would multiply this value by discovered unused IPs to show total estimated savings.
We would be mentioning the same in the readme file. PFB the price documentation page-
https://aws.amazon.com/ec2/pricing/on-demand/#Elastic_IP_Addresses
We have found out one of the API which is also giving the same information based on location, PFB the snap of the api response:
image

Please let us know your suggestions, Thanks

douglaswth:
Using the $0.005 per hour pricing makes sense as long as there are no other discounts involved. However, I don't think you should be considering how many days are remaining in a month for calculating the estimate; the `README.md` already states that the estimate is supposed to be a 30 day estimate, so you should be multiplying the price out to that for the estimate.

KVivek03:
Hi @douglaswth , We tried to research for API which gives the discount details but we are unable to find. We have found out this api -https://docs.aws.amazon.com/aws-cost-management/latest/APIReference/API_GetSavingsPlansCoverage.html, but it seems to be giving on the account level, not specific to idle elastic ips. Please let us know your suggestions. Should we calculate using the $0.005*hour*30 and show per resource and mention in the readme that it is excluding the discounts applied?
cc @nshivappa

douglaswth:
Yes, that seems reasonable for now.

KVivek03:
Hi @douglaswth , We have modified the policy to display savings per resource and total estimated savings using only the per hour IP idle costs for a period of 30 days excluding the discounts, and same we have mentioned in the readme.
Hi @nshivappa , requesting you to kindly review the PR.

cost/aws/unused_ip_addresses/README.md
```diff
@@ -37,7 +37,6 @@ The following policy actions are taken on any resources found to be out of compl
 ## Prerequisites
 
 - This policy uses [credentials](https://docs.rightscale.com/policies/users/guides/credential_management.html) for connecting to the cloud in order to apply this policy you must have a credential registered in the system that is compatible with this policy. If there are no credentials listed when you apply the policy, please contact your cloud admin and ask them to register a credential that is compatible with this policy. The information below should be consulted when creating the credential.
-- billing_center_viewer (note: this role must be applied at the Organization level)
```
douglaswth, line 40: Are you still getting currency information from Optima? If so, you probably need some role in the Flexera One or legacy RightScale platform.

cost/aws/unused_ip_addresses/aws_unused_ip_addresses.pt
```diff
@@ -187,15 +162,27 @@ datasource "ds_get_caller_identity" do
...
+    header "User-Agent", "RS Policies"
+  end
+end
+
+datasource "ds_get_currency_rates" do
+  request do
+    verb "GET"
+    host "api.currencyfreaks.com"
+    path "/latest"
+    query "apikey", "5f996aab23fb4f368057a28bc91b83b1"
```
douglaswth, line 180: Should we really have a hard coded API key here?

nshivappa (commented):
LGTM

nshivappa (dismissed):

KVivek03:
Hi @douglaswth , requesting your approval for the changes. Kindly review the PR Thanks.

cost/aws/unused_ip_addresses/README.md
```diff
@@ -12,8 +12,8 @@ If EIP is not needed, charges can be stopped by releasing the unused EIP.
 
 ### Policy savings details
 
-The policy includes the estimated savings. The estimated savings is recognized if the resource is terminated. Optima is used to receive the estimated savings which is the product of the most recent full day's cost of the resource * 30. The incident header includes the sum of each resource Estimated Monthly Savings in the Incident Header as Total Estimated Monthly Savings.
-If the user is missing the minimum required role of `billing_center_viewer`or if there is no enough data received from Optima to calculate savings, appropriate message is displayed in the incident detail message along with the estimated monthly savings column value as N/A in the incident table.
+The policy includes the estimated savings. The estimated savings is recognized if the resource is terminated. The savings is calculated using the per hour cost of unused IPs for a period of 30 days and is displayed in the Estimated Monthly Savings column and the total estimated sum of all the unused IPs is displayed in the incident detail message.
+Please note that the estimated savings is calculated and displayed in USD($), excluding the discounts.
```
douglaswth, line 16:
```suggestion
Please note that the estimated savings is calculated and displayed in USD($), excluding any discounts.
```

douglaswth (changes requested):
Things are looking good! I just have one suggestion for a wording change in the `README.md`.

douglaswth (approved):

PR #651, opened 2021-01-14
Configurable Optima org for AWS Old Snapshots and AWS Unused Volumes

douglaswth:
### Description

Adds a parameter to the AWS Old Snapshots and AWS Unused Volumes policy templates to override the Flexera One org ID to use when querying Optima for cases when the project is not in the same org where the AWS bill is registered in Optima. The default value for this new parameter is `current` which duplicates the old behavior of using the `rs_org_id` keyword to get the current org ID, so the new behavior will only be triggered if a different org ID value is entered in the parameter. This behaviour is needed in order to get useful cost/savings estimates from these policies for the Flexera FinOps Team's use case where various engineering Flexera One projects that are associated with AWS accounts are not in the same org where we have the AWS bill registered in Optima.

### Issues Resolved

None

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/cost/aws/old_snapshots/CHANGELOG.md
+++ b/cost/aws/old_snapshots/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v2.12
+
+- Add a parameter to override the Flexera One org ID to use when querying Optima for cases when the project is not
+  in the same org where the AWS bill is registered in Optima
+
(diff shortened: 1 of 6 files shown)


nshivappa (commented):
LGTM

nshivappa (approved):

PR #657, opened 2021-02-22
CET-977 Use GitHub GraphQL for enforcing branch protection

mforutan:
### Description

Current issues:

- **GitHub Api Rate limit**: Running current version every hour or less will consume too many calls from the limit.
- **Default Branch**: In the current version is not possible to protect the default branch, it only accept a list of branches which can be different from the real default branch.
- **Minimum configuration**: Current version doesn't change the protection configuration if it doesn't match the parameters, it only checks if there is a branch protection

This PR is fixing the top two issues by using GitHub GraphQL API to grab the data with less api calls (e.g.: 6 calls for flexera with 550 repos) and also to check the default branch instead of using a branch list.

### Issues Resolved

Mentioned above.

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/compliance/github/repository_branch_protection/CHANGELOG.md
+++ b/compliance/github/repository_branch_protection/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Changelog
 
-## v3.0
+## v2.3
 
 - Use GitHub GraphQL API to decrease the number of API calls
-- Remove "branches" parameter and check the protection on the default branch
+- Add option to include default branch regardless of branch selection
 
 ## v2.2
 
--- a/compliance/github/repository_branch_protection/CHANGELOG.md
+++ b/compliance/github/repository_branch_protection/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## v3.0
+
+- Use GitHub GraphQL API to decrease the number of API calls
+- Remove "branches" parameter and check the protection on the default branch
+
 ## v2.2
 
 - Modified escalation label and description for consistency
--- a/compliance/github/repository_branch_protection/repository_branch_protection.pt
+++ b/compliance/github/repository_branch_protection/repository_branch_protection.pt
@@ -5,7 +5,7 @@ long_description ""
 severity "medium"
 category "Compliance"
 info(
-  version: "2.2",
+  version: "3.0",
   provider: "GitHub",
   service: "",
   policy_set: ""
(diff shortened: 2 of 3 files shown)


compliance/github/repository_branch_protection/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v3.0
+
+- Use GitHub GraphQL API to decrease the number of API calls
+- Remove "branches" parameter and check the protection on the default branch
```
douglaswth, line 6: Since this is a Policy Template that is published to customers, I don't know if removing functionality like this is really a good idea.

douglaswth:
Rather than removing the ability to select which branches to enforce protection on and only supporting the default branch, perhaps you could add a new `Yes`/`No` parameter to separately enable checking for the default branch.

compliance/github/repository_branch_protection/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v3.0
```
cdwilhelm, line 3: Please revert the version to v2.3, 3.0 is not a valid version.

compliance/github/repository_branch_protection/repository_branch_protection.pt
```diff
@@ -21,12 +21,6 @@ parameter "param_orgs" do
   label "Organizations on GitHub.com to scan"
 end
 
-parameter "param_branches" do
```
cdwilhelm, line 24: Why are we removing functionality.  since this is a customer public policy we shouldn't remove functionality.

cdwilhelm (changes requested):

compliance/github/repository_branch_protection/CHANGELOG.md
douglaswth: Also, the version would need to be updated within the `info` statement in the Policy Template itself.

compliance/github/repository_branch_protection/CHANGELOG.md
mforutan: I added back the functionality.

compliance/github/repository_branch_protection/CHANGELOG.md
mforutan: I fixed the versions.

compliance/github/repository_branch_protection/repository_branch_protection.pt
mforutan: I added back the functionality.

mforutan:
- Added back the branch selection to match the previous behaviour.
- Added a parameter to include default branch in the check regardless of the branch selection.
- Fix the versions.

bryankaraffa:
If we do not want to push any of these changes, then we can move our modified version of the policy template to our non-customer facing policy template repo: github.com/flexera/devops-policy-templates

compliance/github/repository_branch_protection/repository_branch_protection.pt
mforutan: @cdwilhelm with my latest changes branches can be specified to scan like before, can you please have a look at the new changes when you have a minute?

cdwilhelm (approved):
LGTM

PR #665, opened 2021-03-22
POL-413: Get type from resource, and check whether it supports tags

FabriceScoupeFlexera:
### Description

The resource type can be obtained from the resource as a field, instead of being extracted from the resource id (which doesn't work for types like virtualMachines/extensions).
Not all resources support tags, so a datasource was added to retrieve a CSV table detailing which resource types support tags, and a script to parse that CSV and create a map: resource type -> tag support.
This map is now used to exclude resources that do not support tags from the list of untagged resources.

### Issues Resolved

https://jira.flexera.com/browse/POL-413

### Contribution Check List

- [x] All tests pass.


compliance/azure/azure_untagged_resources/untagged_resources.pt
```diff
@@ -97,20 +97,51 @@ datasource "ds_azure_resources" do
...
+      field "type", jmes_path(col_item,"type")
     end
   end
 end
 
+# See also: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/tag-support
+datasource "ds_tag_support" do
+  request do
+    host "raw.githubusercontent.com"
+    path "/tfitzmac/resource-capabilities/master/tag-support.csv"
```
cdwilhelm, line 109: The csv file should probably live in the flexera/policy_template/data/azure path.  We do this for other things like regions and instance types.

cdwilhelm (changes requested):

cdwilhelm (approved):
lgtm

PR #672, opened 2021-04-12
POL-425: Drop namespace from resource type when looking up API versions.

FabriceScoupeFlexera:
### Description

In the "Tag Resources" action, drop the provider's namespace from the resource type when looking up supported API versions,
since the resource types in the provider object don't have that namespace prepended.

### Issues Resolved

https://jira.flexera.com/browse/POL-425

### Contribution Check List

- [x] All tests pass.


pearson50 (approved):
Looks good to me

String split returns the resource type only and template version is incremented and changed logged

PR #707, opened 2021-06-18
CMLS-11: Make debug a param (off by default), plus other fixes.

FabriceScoupeFlexera:
### Description

Modify templates that log to CM Audit Entries (debug logs); make "$$debug" a parameter that is turned off by default,
since CM is not available in the EU region.

At the same time, fix minor issues (e.g. using hardcoded "optima.rightscale.com" instead of rs_optima_host, etc).

### Issues Resolved

https://jira.flexera.com/browse/CMLS-11

### Contribution Check List

- [x] All tests pass.


pearson50 (approved):
Looks good

Change log and policy version update to match

Logging added 
Host name parameter added where required

PR #742, opened 2021-09-21
IAM Rotate Access Keys - Fix Subject Header

XOmniverse:
### Description
Fixed a small issue in iam_rotate_access_keys.pt where the incident email contained text pasted from another policy that was not changed to fit this policy.


rshade (approved):
LGTM

PR #749, opened 2021-09-28
Pol 478 iam access analyzer enabled

XOmniverse:
### Description

Adds AWS IAM policy to check for regions without an active access analyzer


rshade (dismissed):
LGTM

PR #763, opened 2021-10-19
Adds an AWS CIS policy for: Storage 2.1.3 Ensure MFA Delete is enabled on S3 buckets

XOmniverse:
### Description

Adds an AWS CIS policy for: Storage 2.1.3 Ensure MFA Delete is enabled on S3 buckets


lawrethao (dismissed):
Changes look good, script looks to do as defined.

PR #779, opened 2021-11-08
Pol 508 add cost to google recommender

XOmniverse:
### Description

Adds functionality to the Google Recommender policy. Specifically, it adds Cost and Currency fields to the incident output so that the user knows how much money is at stake, and it allows (optionally) the policy to filter by specific projects if desired by the end user. 

I have tested these changes and found no errors or problems.


rshade (approved):
LGTM

PR #780, opened 2021-11-10
Google Cost Recommender: Added cost duration in months to incident output

XOmniverse:
### Description

Added cost duration in months to incident output. Also removed the negative value from the cost in order to make the output more clear to the user.

Tested and confirmed it works as expected.

### DIFF
--- a/cost/google/recommender/CHANGELOG.md
+++ b/cost/google/recommender/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.2
+
+- Added cost to incident output
+
 ## v2.1
 
 - Updating the short description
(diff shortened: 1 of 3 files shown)


cdwilhelm (changes requested):
Can we change cost to Monthly Savings, and include the currency symbol to the Monthly Savings like we do in other Policies.  Using a column for currency takes value away from more important columns.  Also the UI only shows the first 5 columns.  The most important columns should be showing first.

XOmniverse:
> Can we change cost to Monthly Savings, and include the currency symbol to the Monthly Savings like we do in other Policies. Using a column for currency takes value away from more important columns. Also the UI only shows the first 5 columns. The most important columns should be showing first.

I made most of these changes. I used "Savings" instead of "Monthly Savings" since it is not a given that it will always be for a single month. The column for "Savings Duration (Months)" should cover the amount of time.

jez-chaplin (approved):

rshade:
closing this as we are going to do a bunch of discrete policies.

PR #781, opened 2021-11-10
Policies for Azure AKS Autoscaling

XOmniverse:
This adds two policies; one for checking whether any AKS clusters have nodepools without autoscaling enabled, and another for checking whether any AKS clusters have nodepools without zero autoscaling (autoscaling with a minimum node count of 0) enabled.

Both policies have been tested.


operational/azure/aks_nodepools_without_autoscaling/aks_nodepools_without_autoscaling.pt
```diff
@@ -0,0 +1,217 @@
...
+  type "list"
+  label "Email addresses of the recipients you wish to notify."
+end
+
+parameter "param_exclusion_tags" do
+  category "User Inputs"
+  label "Tags To Ignore"
+  description "The policy will ignore any clusters with the specified tags. Only supply the tag keys; the policy assumes that the tag value is irrelevant."
+  type "list"
+  default []
```
rshade, line 30: Can you remove this?

operational/azure/aks_nodepools_without_zero_autoscaling/aks_nodepools_without_zero_autoscaling.pt
```diff
@@ -0,0 +1,226 @@
...
+  type "list"
+  label "Email addresses of the recipients you wish to notify."
+end
+
+parameter "param_exclusion_tags" do
+  category "User Inputs"
+  label "Tags To Ignore"
+  description "The policy will ignore any clusters with the specified tags. Only supply the tag keys; the policy assumes that the tag value is irrelevant."
+  type "list"
+  default []
```
rshade, line 30: Can you remove this?

rshade (changes requested):

operational/azure/aks_nodepools_without_autoscaling/aks_nodepools_without_autoscaling.pt
XOmniverse: @rshade The entire thing or just the "default []"? If the former, I'm happy to do so, but I'd be curious why since my understanding is that we generally want to give the option to filter for tags.

rshade (approved):
LGTM

PR #802, opened 2021-11-17
POL-519 - AWS CIS Ensure that Object-level logging for read and write events is enabled for S3  bucket 

XOmniverse:
### Description
Adds policy that covers the following:

AWS CIS 3.10 Ensure that Object-level logging for write events is enabled for S3 bucket.

AWS CIS 3.11 Ensure that Object-level logging for read events is enabled for S3 bucket.


rshade (approved):
LGTM

PR #807, opened 2021-11-22
cost/azure/old_snapshots: Added subscription filter option and ability to specify Azure API endpoint

XOmniverse:
### Description

cost/azure/old_snapshots: Added subscription filter option and ability to specify Azure API endpoint


rshade (approved):
LGTM

PR #820, opened 2021-11-22
OOTB Policies: Add Subscription Filtering To cost/azure/reserved_instances/mca_recommendations

XOmniverse:
### Description

OOTB Policies: Add Subscription Filtering To cost/azure/reserved_instances/mca_recommendations


rshade (approved):
LGTM

PR #839, opened 2021-11-24
OOTB Policies: Add Subscription Filtering To compliance/azure/compliance_score

XOmniverse:
### Description

OOTB Policies: Add Subscription Filtering To compliance/azure/compliance_score

(Also added 400 to the ignore_status on line 98 since this might be a valid API response)


rshade (approved):
LGTM

PR #849, opened 2021-12-03
POL-521 - Make AWS policies ready for spend recommendations

lawrethao:
### Description

This change updates certain cost policies to make them ready for spend recommendations 

### Issues Resolved

these changes resolve issues with field mappings in certain cost policy templates

### Contribution Check List

- [ ] All tests pass.


cost/aws/old_snapshots/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.17
+
+- added changes from [wiki](https://confluence.flexera.com/pages/viewpage.action?spaceKey=rightscale&title=Usage+Recommendations%3A+Waste+reduction)
```
douglaswth, line 5: This is an internal document and should not be referenced in a public ChangeLog.

cost/aws/unused_ip_addresses/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.16
+
+- added changes from [wiki](https://confluence.flexera.com/pages/viewpage.action?spaceKey=rightscale&title=Usage+Recommendations%3A+Waste+reduction)
```
douglaswth, line 5: Same as above.

cost/aws/unused_rds/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.11
+
+- added changes from [wiki](https://confluence.flexera.com/pages/viewpage.action?spaceKey=rightscale&title=Usage+Recommendations%3A+Waste+reduction)
```
douglaswth, line 5: Same as above.

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -251,7 +251,7 @@ datasource "ds_get_snapshot_ami" do
       field "imageId", xpath(col_item, "imageId")
       field "region", val(iter_item, "region")
       field "snapshotId", val(iter_item, "snapshotId")
-      field "tagKeyValue", val(iter_item, "tagKeyValue")
+      field "tags", val(iter_item, "tags")
```
douglaswth, line 254: Have you tested that the `tags` field shows up in the XML with exactly the same format as `tagKeyValue` did? It seems odd that AWS would have changed the format of the `2016-11-15` version of the `DescribeImages` response.

cost/aws/idle_compute_instances/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.14
+
+- added changes from [wiki](https://confluence.flexera.com/pages/viewpage.action?spaceKey=rightscale&title=Usage+Recommendations%3A+Waste+reduction)
```
douglaswth, line 5: Same as below.

douglaswth (changes requested):

douglaswth:
Looks like all the changelog entries do not have matching version bumps in the `info` blocks of the actual Policy Templates as well.

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -1,12 +1,12 @@
-name "AWS Old Snapshots"
+name "AWS Old Snapshots test"
```
jim-slattery-rs, line 1: Committed accidentally?
```suggestion
name "AWS Old Snapshots"
```

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -655,7 +644,7 @@ EOS
       field "savings" do
         label "Estimated Monthly Savings"
       end
```
jim-slattery-rs, line 646: For these five AWS usage recommendation policies, could we further refine the "savings" field content?

In the incident violation_data, I noticed the `savings` value included currency info as well as an amount: [sample_responses/org_27018/incident_aws_unused_volumes.json#L310](https://github.com/flexera/optima-recommendations/blob/3da2ff02a004487a50396f2b6aa4401b7358350c/front_service/services/incident/testing/sample_responses/org_27018/incident_aws_unused_volumes.json#L310)
```
"savings": "US$ 0.83",
```
Could we instead maybe separate that, so it would be:
```
"savings": 0.83,
"savings_currency": "US$",
```
That way, we can scrape the "savings" amount for recommendations in a standard way, without any string processing.

douglaswth: While I think making this change is a good idea, we should follow semantic versioning and bump the major version instead of the minor version since it could be considered a backward incompatible change. I believe this would probably result in most of the policies becoming `3.0` versions.

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -625,7 +615,6 @@ policy "pol_utilization" do
     summary_template "AWS Account ID: {{with index data.result 0}}{{ .accountId }}{{end}} - {{ len data.result }} rows containing AWS old snapshot data"
     detail_template <<-EOS
 The following {{ len data.result }} old snapshots, for AWS Account: {{with index data.result 0}}{{ .accountId }}{{end}}, have exceeded the specified age of: {{ parameters.snapshot_age }} days old.
-
```
douglaswth, line 628: This extra newline was actually meaningful in the resulting markdown from the template, please add it back.

douglaswth (changes requested):

cost/aws/idle_compute_instances/idle_compute_instances.pt
```diff
@@ -795,12 +795,18 @@ EOS
       field "cpu_average" do
         label "CPU Average %"
       end
+      field "instanceType" do
+        label "Instance Type"
+      end
```
jim-slattery-rs, line 800: Curt would like the recommendations to have a generic "Resource Type" field, so how about we scrape that from a standard `resourceType` field in these usage recommendations?

Currently, they may have an `instanceType` or `volumeType` field.
We can either rename those to `resourceType`, while still keeping the field "label" as "Instance Type" or "Volume Type".
Or we can add `resourceType` as an extra field, in addition to `instanceType` or `volumeType`.

cost/aws/idle_compute_instances/idle_compute_instances.pt
```diff
@@ -700,7 +700,7 @@ script "js_instance_cost_mapping", type:"javascript" do
         idle_instance.push({
           "accountId": ds_get_caller_identity[0]['account']
```
jim-slattery-rs, line 701: Since we're standardizing these fields, should we use `accountID` in these policies, matching the capitalization in `resourceID`?
```suggestion
          "accountID": ds_get_caller_identity[0]['account']
```

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -446,28 +440,29 @@ script "js_snapshots_cost_mapping", type: "javascript" do
         accountId: ds_get_caller_identity[0]['account'],
         snapshotId : snapshot['snapshotId'],
         region : snapshot['region'],
-        tagKeyValue : snapshot['tagKeyValue'],
+        tags : snapshot['tagKeyValue'],
         daysOld : snapshot['daysOld'],
         imageId : snapshot['imageId'],
         snapshotImageIdFormat : snapshot['snapshotImageIdFormat'],
-        savings : cur + ' '+formatNumber(monthly_savings.toFixed(3), separator),
+        savings : monthly_savings.toFixed(3),
```
jim-slattery-rs, line 447: I'm seeing the incident having a string `savings` field instead of a number:
```
      "savings": "0.237",
```
In these policies, can we keep it as a number instead?
```
      "savings": 0.237,
```

Maybe we could skip calling `toFixed`?
```suggestion
        savings : monthly_savings,
```
 Or otherwise parse the string back to a number?:
```suggestion
        savings : Number(monthly_savings.toFixed(3)),
```

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
rshade: Do we even need the currency anymore or can recommendations just pull it directly?

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -443,31 +437,32 @@ script "js_snapshots_cost_mapping", type: "javascript" do
...
+        accountID: ds_get_caller_identity[0]['account'],
         snapshotId : snapshot['snapshotId'],
         region : snapshot['region'],
-        tagKeyValue : snapshot['tagKeyValue'],
+        tags : snapshot['tagKeyValue'],
         daysOld : snapshot['daysOld'],
         imageId : snapshot['imageId'],
         snapshotImageIdFormat : snapshot['snapshotImageIdFormat'],
-        savings : cur + ' '+formatNumber(monthly_savings.toFixed(3), separator),
+        savings : Number(monthly_savings.toFixed(3)),
```
jim-slattery-rs, line 447: Ah, looks like the `Number` function isn't available: the savings in the incident are now all showing up as zero.

Maybe we can just remove `toFixed` altogether and use the raw number?
```suggestion
        savings : monthly_savings,
```

cost/aws/idle_compute_instances/idle_compute_instances.pt
```diff
@@ -713,13 +713,14 @@ script "js_instance_cost_mapping", type:"javascript" do
           "mem_average": instance["mem_average"],
           "mem_maximum_value": instance["mem_maximum_value"],
           "mem_average_value": instance["mem_average_value"]
-          "savings": cur + ' '+formatNumber(monthly_savings.toFixed(3), separator)
+          "savings": Number(monthly_savings.toFixed(3)),
```
jim-slattery-rs, line 716: Since `Number` isn't available, maybe just remove `toFixed`?
```suggestion
          "savings": monthly_savings,
```

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
jim-slattery-rs: I think it's great if the policy annotates the savings number with whatever currency information it happened to use.

Today, Optima just has a [currency_setting](https://reference.rightscale.com/bill_analysis/#/currency_setting/currency_setting_show) that is set for the entire org. 
But to really make sense of multiple currencies in the same org, we'll need to ensure all of our cost sources indicate what currency unit they're using, so we can avoid summing up costs with disparate currencies.

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
jim-slattery-rs: Or `parseFloat` would probably work as an alternative to `Number`, at least I've seen that function called in these policies.e

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
rshade: yeah, but the policy just pulls in currency_setting to display it back to you: https://github.com/flexera-public/policy_templates/blob/57b01c17f8d486e3579041c646b0b34f6dd1d68c/cost/aws/idle_compute_instances/idle_compute_instances.pt#L224

jim-slattery-rs: Good point! 😄 
I guess that currency info could still be useful to those viewing the raw incident?

cost/aws/unused_rds/unused_rds.pt
```diff
@@ -329,21 +329,38 @@ end
...
   for ( n=0; n < ds_rds_instance_set.length; n++) {
     var instance = ds_rds_instance_set[n]
     // get tags for instance
     var instances_w_tags = _.where(ds_list_tags_for_resource, {instanceId: instance.instanceId})
     var instance_tags = {}
-    var tag_set = []
+    var tags = []
     _.each(instances_w_tags, function(instance){
-      tag_set.push(instance.key+':'+instance.value)
+      tags.push(instance.key+':'+instance.value)
```
jim-slattery-rs, line 361: For recommendations to scrape the tags, we'll need them to use an `=` separator between key and value:
```suggestion
      tags.push(instance.key+'='+instance.value)
```

cost/aws/unused_rds/unused_rds.pt
```diff
@@ -509,16 +529,21 @@ script "js_instance_cost_mapping", type:"javascript" do
       }
       var total_savings=0;
       _.each(instance_list, function(instance){
+        console.log(instance)
```
jim-slattery-rs, line 532: Not sure if we want to remove this `console.log` line now?

cost/aws/idle_compute_instances/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v3.0
+
+- applying data normalization updates for spend recommendations api
```
jim-slattery-rs, line 5: Do we want to add more specifics to summarize the changes in each of these policies?
For example:
```suggestion
- Normalizing fields for recommendations:
  - Renamed 'tag_set' to 'tags'
  - Renamed 'accountId' to 'accountID'
  - Renamed 'id' to 'resourceID'
  - Renamed 'instanceType' to 'resourceType'
  - Changed 'savings' to be a plain number
  - Added 'savings_currency' to record the currency symbol.
  - Added a 'service' field, hardcoded to "EC2"
```

cost/aws/unused_rds/unused_rds.pt
lawrethao: thanks jim! good catch.

cost/aws/idle_compute_instances/CHANGELOG.md
lawrethao: yes, curt actually mentioned being very specific for the 3.0. I also missed updating this to be more specific. again thanks for the catch. I've updated these requests.

cost/aws/idle_compute_instances/idle_compute_instances.pt
```diff
@@ -713,17 +726,20 @@ script "js_instance_cost_mapping", type:"javascript" do
           "mem_average": instance["mem_average"],
           "mem_maximum_value": instance["mem_maximum_value"],
           "mem_average_value": instance["mem_average_value"]
-          "savings": cur + ' '+formatNumber(monthly_savings.toFixed(3), separator)
+          "savings": parseFloat(monthly_savings.toFixed(3)),
```
douglaswth, line 729: Rather than rendering to a string and then parsing it back to a number, it would probably be better to do something like:

```javascript
Math.round(monthly_savings * 1000) / 1000;
```

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
```diff
@@ -443,31 +437,34 @@ script "js_snapshots_cost_mapping", type: "javascript" do
...
         snapshotId : snapshot['snapshotId'],
         region : snapshot['region'],
-        tagKeyValue : snapshot['tagKeyValue'],
+        tags : snapshot['tagKeyValue'],
         daysOld : snapshot['daysOld'],
         imageId : snapshot['imageId'],
         snapshotImageIdFormat : snapshot['snapshotImageIdFormat'],
-        savings : cur + ' '+formatNumber(monthly_savings.toFixed(3), separator),
-        volumeSize : snapshot['volumeSize']
+        savings : parseFloat(monthly_savings.toFixed(3)),
```
douglaswth, line 447: Rather than rendering to a string and then parsing it back to a number, it would probably be better to do something like:

```javascript
Math.round(monthly_savings * 1000) / 1000;
```

douglaswth (changes requested):

cost/aws/old_snapshots/aws_delete_old_snapshots.pt
lawrethao: good catch with it no longer being string. added these and standardized it at 3 decimals for all of the templates

cost/aws/unused_volumes/aws_delete_unused_volumes.pt
```diff
@@ -709,7 +700,7 @@ script "js_volume_cost_mapping", type:"javascript" do
   result={
     "unused_volumes": unused_volumes,
     "message": message,
-    "accountId": ds_get_caller_identity[0]['account']
+    "accountID": String(ds_get_caller_identity[0]['account'])
```
jim-slattery-rs, line 703: Checked with Ren, this `String()` call was not needed in the end, and will be removed.

douglaswth (approved):

PR #948, opened 2022-03-23
ITAM Policy: VMs Without Host ID

XOmniverse:
ITAM Policy to find VMs without a Host ID. Policy has been tested and works as expected.


rshade (changes requested):
Can you add this to the root readme?

rshade (approved):
LGTM

PR #968, opened 2022-03-28
Azure CIS 2.11 Ensure That Auto provisioning of 'Log Analytics agent for Azure VMs' is Set to 'On'

XOmniverse:
### Description

Azure CIS 2.11 Ensure That Auto provisioning of 'Log Analytics agent for Azure VMs' is Set to 'On'


rshade (approved):
LGTM

PR #978, opened 2022-03-30
Azure CIS 2.14 Ensure That 'Notify about alerts with the following severity' is Set to 'High'

XOmniverse:
Azure CIS 2.14 Ensure That 'Notify about alerts with the following severity' is Set to 'High'

Tested and works


rshade (approved):
LGTM

PR #988, opened 2022-04-06
3.5 Ensure that 'Public access level' is set to Private for blob containers 

XOmniverse:
3.5 Ensure that 'Public access level' is set to Private for blob containers 

Tested and works fine.


rshade (approved):
LGTM

PR #999, opened 2022-04-06
Azure CIS 3.10/3.11 Ensure Storage logging is Enabled for Blob/Table Service for 'Read', 'Write', and 'Delete' requests

XOmniverse:
Azure CIS 3.10/3.11 Ensure Storage logging is Enabled for Blob/Table Service for 'Read', 'Write', and 'Delete' requests

tested and working

(Two policies in one pull request because they are extremely similar to one another)


rshade (approved):
LGTM

PR #1001, opened 2022-04-07
POL-611 Invert AWS Disallowed Regions Policy

XOmniverse:
This is a modification of the AWS Disallowed Regions Policy. The user now has to specify a list of regions to allow rather than a list of regions to disallow. There's also been some general cleanup to bring the policy more in line with our current best practices.

Tested and works as expected.


rshade (approved):
LGTM

PR #1014, opened 2022-04-12
Azure CIS 4.1.3 Ensure that 'Auditing' Retention is 'greater than 90 days' 

XOmniverse:
4.1.3 Ensure that 'Auditing' Retention is 'greater than 90 days' 

Tested and works.

(Note: Despite the title in the CIS document, the actual description in the document is "Ensure that RetentionInDays is set to more than or equal to 90" rather than "greater than 90 days". I opted for the former and named the policy accordingly)


taterdog590 (dismissed):
Approved

bryankaraffa (approved):

PR #1065, opened 2022-06-15
Fix: Add `auth: $$auth_rs` for authentication to policy sync template

bryankaraffa:
### Description

Adds `auth: $$auth_rs` to the Flexera API requests in CWF definition.  Currently actions are receiving unauthorized and this is the reason EU Policies are not being published/updated as expected.


beagledeltasix (approved):
Approved

PR #1066, opened 2022-06-15
Fix: Disable random branch suffix on Actions Workflow

bryankaraffa:
### Description

Modifies the GitHub Actions workflow that creates a "Release Candidate" Pull Request to sync changes from `master` branch into `EU_Policies` branch.

Disabling random branch creation will reduce the amount of Pull Requests that are created by this automation and disable the amount of manual cleanup necessary.  Using alternative branching strategies like random is also [not recommended by the actions developer](https://github.com/peter-evans/create-pull-request#alternative-strategy---always-create-a-new-pull-request-branch).

PR #1069, opened 2022-06-15
Revert #1066

bryankaraffa:
Reverts changes from #1066 .  This is required because we can't resolve conflicts with a `master` -> `EU_Policies` PR (Example in https://github.com/flexera-public/policy_templates/pull/1067 )


beagledeltasix (approved):
Reverting change.. Approved

PR #1083, opened 2022-07-18
POL-670 AWS Idle Compute Instance - Bulk API Requests

XOmniverse:
### Description
Fundamental refactor of the AWS Idle Compute Instance policy to use a more efficient API for gathering cloud data.


bryankaraffa (approved):
👍 

Only question/discussion before release would be the release type [minor release -- `v3.5`].  We added a new parameter, and new prams are breaking changes for users applying the policy via API because the new param would need to be added to their http request.  

I feel a major release [`v4.0`] would more closely aligned with [semver](https://semver.org/) if we are following that release strategy, but @cdwilhelm mentioned that we have used minor releases for similar changes in the past..

cdwilhelm (approved):
LGTM

PR #1092, opened 2022-07-26
Throttling - Azure Storage Accounts without Lifecycle Management Policies

XOmniverse:
### Description

Added functionality to allow requests to be delayed to avoid having the Azure API throttle the user. I've tested the policy and it works as expected. No other functionality was added.

(Change was made to accommodate PepsiCo but I saw no reason not to just make it available to everyone)


nia-vf1 (approved):
LGTM

PR #1104, opened 2022-08-18
feat: DescribeRegions filter out not-opted-in

bryankaraffa:
### Description

[`ec2:DescribeRegions`](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRegions.html) returns all regions by default -- even those not opted into yet.  New regions by default are not opted into as well, so this combination has resulted increased policy template errors when running without specifying region whitelist [`param_allowed_regions`].  

This change filters only regions that are `opted-in` or `opt-in-not-required` in the current account

### Contribution Check List

- [x] All tests pass.

### DIFF
--- a/compliance/aws/disallowed_regions/CHANGELOG.md
+++ b/compliance/aws/disallowed_regions/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.7
+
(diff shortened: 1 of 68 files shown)


taterdog590 (approved):
LGTM

PR #1114, opened 2022-09-12
Tag cardinality

XOmniverse:
Adds policies for AWS and Azure to generate tag cardinality reports. These policies have been tested.

Does not currently include GCP; this is TBD due to difficulties with the GCP API.


taterdog590 (approved):
LGTM

PR #1129, opened 2022-11-18
POL-731 Update Azure Unused IP Addresses

nia-vf1:
### Description

Bug fixes - 

- added `ignore_status` for 400 and 404 error codes on Azure APIs to prevent policy from failing when iterating over certain subscriptions
- added condition to handle `resourceType` field where it is equal to null, again to prevent policy from failing

### Issues Resolved

### Contribution Check List

- [ ] All tests pass.


cost/azure/unused_ip_addresses/azure_unused_ip_addresses.pt
```diff
@@ -283,7 +284,7 @@ script "js_get_costs", type:"javascript" do
         "User-Agent": "RS Policies",
         "Api-Version": "1.0"
       },
-      ignore_status: [400]
+      ignore_status: [400, 404]
```
bryankaraffa, line 287: Why is 400 being ignored again?

taterdog590: 400's need to be ignored to avoid legacy Azure AD subscriptions throwing 400 error on requesting any resource via api. MSFT has stated this is not exposed on newer tenants, but no way to filter out these older subscriptions on customer with legacy EA / Azure setups. We actually need to add this to all azure policy, in all sections that check resources.

bryankaraffa (approved):

PR #1143, opened 2022-12-15
HCLIND-1 AWS reserved instance option to exclude compute

degarajesh:
### Description

AWS Reserved Instance Recommendation: Service option to exclude compute

### Issues Resolved

Add new service selection "All Except EC2"

### Contribution Check List

- [ ] All tests pass.


lawrethao (changes requested):
dont forget to update version in the .pt file info block: 
info(
  version: "2.15",
  provider: "AWS",
  service: "EC2",
  policy_set: "Reserved Instance"
)

add a small blurb in the readme on reason for the need for all but EC2 as an option from the jira ticket.

lawrethao (approved):
LGTM

degarajesh:
@lawrethao I added a fix here. Please review whenever you get time.

lawrethao (approved):

cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt
```diff
@@ -29,9 +29,9 @@ end
 parameter "param_service" do
   category "RI"
   label "Service"
-  default "All"
+  default "All Except EC2"
```
bryankaraffa, line 32: Changing default value for an existing param is a breaking change.  Also, should this new value be the default?

cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt
```diff
@@ -201,11 +201,17 @@ script "js_ri_types", type: "javascript" do
...
-      var services = ["ElastiCache","Elastic Compute Cloud (EC2)","Elasticsearch Service (ES)","Redshift","Relational Database Service (RDS)"]
       _.each(services, function(service){
         result.push({"service":service})
       })
+    } else if (param_service === "All Except EC2"){
+      services.forEach(function(service) {
+        if (service !== "Elastic Compute Cloud (EC2)") {
+          result.push({"service":service})
+        }
+    });
```
bryankaraffa, line 214: Seems like it would be better to add a new param (`param_services_exclude`) instead so it's more extendable (?)

bryankaraffa (commented):
Few comments.. Something that would help is clarity on the use-case these changes are associated with?

cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt
degarajesh: good thought!, As per ticket, "add new selection to service input"

https://flexera.atlassian.net/browse/HCLIND-1

I will leave it to @cdwilhelm

cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt
degarajesh: answered in the below comment.

cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt
cdwilhelm: @degarajesh    The goal is to exclude services included in the Savings Plans so the savings are not duplicated.  There are many ways to solve this.  
I can see more services covered by Savings Plans in the future, such as Databases. Bryan raises a good point to make it extensible.  
@bryankaraffa   how do you see a user of this policy configuring it to exclude EC2 or other services?  What options  would they select in the param_services and param_services_excluded

degarajesh:
In order to fix build issues, will be creating new pull requests from the flexera_templates not from the forked repo.

taniaSol (commented):
@bryankaraffa @cdwilhelm I remember this suggestion being made. At some point we decided to add "All Except EC2" as was asked to get this quickly done and create an additional ticket for the enhancement @bryankaraffa suggested. 
There are pros and cons in both approaches:
"**All Except EC2**": easier to use as it's just a choice from the drop down. Less flexible, if customer wants to get a subset of services -> need to apply multiple services.
**Include/Exclude**: user will need to type service names, but it's provides ultimate flexibility.

Now that due to GitHub issue we weren't able to finish work quickly, I am inclined to take Bryan's idea. We could pre-populate include list with all services by default to save on user typing

PR #1164, opened 2023-01-16
HCLIND-1 AWS reserved instance option to exclude compute

degarajesh:
### Description

AWS reserved instance option to exclude compute

### Issues Resolved

### Contribution Check List

- [ ] All tests pass.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Editor-based HTTP Client requests
-/httpRequests/
-# Datasource local storage ignored files
-/dataSources/
-/dataSources.local.xml
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,6 +1,0 @@
-<component name="InspectionProjectProfileManager">
-  <profile version="1.0">
-    <option name="myName" value="Project Default" />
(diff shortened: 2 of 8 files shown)


taniaSol (commented):
in README, we should remove "to avoid creating exaggerated potential savings" as this would happen if this policy is used in conjunction with "AWS Savings Plan Recommendations" policy.

degarajesh:
> in README, we should remove "to avoid creating exaggerated potential savings" as this would happen if this policy is used in conjunction with "AWS Savings Plan Recommendations" policy.

@taniaSol Updated can you please have a look ?

bryankaraffa:
@taniaSol  I did a review in another PR for this change, want to bubble them up here so they don't get lost
https://github.com/flexera-public/policy_templates/pull/1143#discussion_r1060764854
https://github.com/flexera-public/policy_templates/pull/1143#discussion_r1060767642

And bringing an answer @cdwilhelm 's question in that PR to this PR as well:
> @degarajesh The goal is to exclude services included in the Savings Plans so the savings are not duplicated. There are many ways to solve this. I can see more services covered by Savings Plans in the future, such as Databases. Bryan raises a good point to make it extensible. @bryankaraffa how do you see a user of this policy configuring it to exclude EC2 or other services? What options would they select in the param_services and param_services_excluded
 
There could be two params, both by default are empty lists:
`param_services_include=[]`
`param_services_exclude=[]`
In the PT, if length(`param_services_include`) == 0, then use with list of services defined in PT ("all services").  If param_services_include is not empty, then the list from that param value.
With that services list defined, remove any from the list that exist in the `param_services_exclude`.

This would enable a customer to exclude one (or more) service -- by leaving include empty and providing value in exclude.  A customer could also include only one (or more service) -- by defining include and leaving exclude empty.

taniaSol (commented):
@bryankaraffa @degarajesh I spoke with Doug and he believes 99.9% of use cases will be choosing all services except EC2. Let's finish this one in line with the original ask and we can improve later by adding more flexibility as Bryan suggested.

bryankaraffa:
@taniaSol  -- Feel free to give you're formal approval on this if you feel it's good enough

bryankaraffa (commented):
.

taniaSol (approved):

PR #1168, opened 2023-02-01
fix tested

XOmniverse:
I discovered an issue where the tag key filtering was not working correctly. It seems like there are edge cases where, depending on the credential permissions and level of access, the Azure API returns values for the "tags" field that the original policy did not handle correctly, resulting in all VMs being filtered, and thus no results being generated, even if the tag filter parameter was left blank.

I reworked the relevant script block so that it no longer has this issue, and I've tested it in our testing environment to ensure that the policy works as expected both when filtering by tag key and when leaving the field blank.


taterdog590 (approved):
LGTM

PR #1174, opened 2023-02-09
For for Graph when using "Week"

XOmniverse:
Fixed issue with incorrectly-rendered graph when selecting "Week" for the Billing Term parameter


taterdog590 (approved):
LGTM, quite the fix!

PR #1205, opened 2023-03-06
POL-779 Add RBD Generation Policies

XOmniverse:
This request will add policies for automatically generating RBDs for account tags/labels. The publish has been set to 'false' for all policies as these are primarily used by our solutions architects rather than customers directly. 

These policies have been widely used by the above team for months, so it is fair to say that they are well tested and work as expected.


automation/aws/aws_rbd_from_tag/aws_rbd_from_tag.pt
```diff
@@ -0,0 +1,265 @@
+name "AWS Rule-Based Dimension From Account Tags"
+rs_pt_ver 20180301
+type "policy"
+short_description "Creates and/or updates individual Rule-Based Dimensions based on AWS Account tags. See the [README](https://github.com/flexera-public/policy_templates/tree/master/automation/aws/aws_rbd_from_tag) and [docs.flexera.com/flexera/EN/Automation](https://docs.flexera.com/flexera/EN/Automation/AutomationGS.htm) to learn more."
+long_description ""
+severity "low"
+category "Cost"
+default_frequency "daily"
+info(
+  version: "2.0",
```
bryankaraffa, line 10: `v0.1`

Going to start using Semantic Versioning for new policies [Teams thread](https://teams.microsoft.com/l/message/19:833373548e104af2a20b0216eda1ba7b@thread.skype/1677611086855?tenantId=91034d23-0b63-4943-b138-367d4dfac252&groupId=fb250818-e040-4a26-b207-61c3cd99fd6e&parentMessageId=1677610206124&teamName=Team%20Flexera%20One&channelName=Policy%20Support%20and%20Questions&createdTime=1677611086855&allowXTenantAccess=false)

automation/aws/aws_rbd_from_tag/aws_rbd_from_tag.pt
```diff
@@ -0,0 +1,265 @@
...
+      "Api-Version": "1.0",
+      "content-type": "application/json"
+    },
+    body_fields: {
+      "name": name,
+      "dated_rules": dated_rules
+    }
+  }
+EOS
+end
```
bryankaraffa, line 253: What was the reasoning for using DS to handle applying the resulting RBD rules ?

bryankaraffa (commented):
Good initial revision.  

Some more comments in the policy templates to explain what each step is doing and why we need to do it would be helpful -- unclear to me why we're using datasource for the apply RBD action instead of CWF and that would be helpful.

Can these Policy Templates be used together?  i.e. if user has AWS, Google and Azure cloud accounts all tagged with "Team" and "Account Type", would the policies aggregate the rules from multiple cloud into a rule for `rbd_environment` (?)

Open to discuss here or over a screenshare

automation/aws/aws_rbd_from_tag/aws_rbd_from_tag.pt
bryankaraffa: bump @XOmniverse ^^  If these are being used we should consider merging them but wanted to sync again on this question firs

PR #1229, opened 2023-04-05
Add Child Policy Code to AWS Object Storage Optimization Policy

XOmniverse:
This just adds the additional code needed for this policy to work as a child policy.

Functionality as a regular policy is unchanged apart from the addition of a Flexera credential, since this is needed for the child policy functionality but was not present in the original policy.

The version is also bumped from 2.9 to 3.0, so even if the above is considered a "breaking change" (I imagine most customers have a Flexera credential at this point), the major version number also changed.

### DIFF
--- a/cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
+++ b/cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
@@ -7,7 +7,7 @@ category "Cost"
 severity "low"
 default_frequency "daily"
 info(
-  version: "2.10",
+  version: "3.0",
   provider: "AWS",
   service: "S3",
   policy_set: "Object Store Optimization"
--- a/cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
+++ b/cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
@@ -7,7 +7,7 @@ category "Cost"
 severity "low"
 default_frequency "daily"
 info(
-  version: "3.0",
+  version: "2.10",
   provider: "AWS",
   service: "S3",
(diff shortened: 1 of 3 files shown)


cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
```diff
@@ -7,7 +7,7 @@ category "Cost"
 severity "low"
 default_frequency "daily"
 info(
-  version: "2.9",
+  version: "3.0",
```
bryankaraffa, line 10: This is a minor version change so would be `2.10`
> MAJOR version when you make incompatible API changes
> MINOR version when you add functionality in a backwards compatible manner
https://semver.org/#summary

bryankaraffa (changes requested):
Forgot the changes to the "first" datasource;
https://github.com/flexera-public/policy_templates/blob/POL-649/meta_policies/README_META_POLICIES.md#identify-first-datasource-and-add-header

bryankaraffa (changes requested):
Also missing changes to `check` statement:
https://github.com/flexera-public/policy_templates/blob/POL-649/meta_policies/README_META_POLICIES.md#modify-policy-validation-check-statements

That [Meta Policy Development](https://github.com/flexera-public/policy_templates/blob/POL-649/meta_policies/README_META_POLICIES.md#meta-policy-development) Section in the README has all changes necessary for creating a net new Child and Meta Parent PT

XOmniverse:
All of the mentioned problems have been resolved:

* Version changed to 2.10
* Check statement modified
* Header added to first datasource that executes. I also moved this datasource to the top of the policy to avoid confusion.

PR #1230, opened 2023-04-05
Meta Policy Support for AWS Object Storage Optimization

XOmniverse:
Adds a meta policy for AWS Object Storage Optimization

Intended to function correctly once the following pull request is approved:
https://github.com/flexera-public/policy_templates/pull/1229


bryankaraffa (commented):
Meta Parent seems good -- I would prefer test a modified version of it, that uses the modified Child Policy Template (version uploaded, not from catalog) before we merge this.. Once changes requested to child PT in https://github.com/flexera-public/policy_templates/pull/1229 are done happy to make those edits locally and help test this meta parent

bryankaraffa (approved):

PR #1248, opened 2023-04-19
AWS Unused Volumes: Minor Incident Fix

XOmniverse:
### Description

Changed "Id" field in incident output to "ID" to match other policies and ensure proper reporting. 

### Issues Resolved

I suspect this is causing the "ids" field to not appear in the action status for actions taken by this policy, which can impact someone's ability to use the policy API to see which specific resources were actioned.

The change is so minor, it is very unlikely to break the policy and it matches code in other policies.

### Contribution Check List

There is no new functionality of note.

### DIFF
--- a/cost/aws/unused_volumes/CHANGELOG.md
+++ b/cost/aws/unused_volumes/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.1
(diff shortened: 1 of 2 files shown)


taterdog590 (approved):
Great change!

PR #1256, opened 2023-04-26
Adds Azure MySQL Instance Size Data (Related to POL-793)

XOmniverse:
### Description

Adds Azure MySQL Instance Size Data for use with policies that need it. Data was produced manually, so long term we probably want to rely on an API to gather this data.


nia-vf1 (approved):
LGTM

PR #1257, opened 2023-04-27
FOPTS-717 Remove limit for the threshold parameter for Budget Alert Policy

night-codes:
### Description

Threshold currently allows up to 100%. This parameter doesn’t accept an input that is above 100% (i.e. max budget). For many customers they are comfortable with a budget over run of up to 105% or 110%. Suggestion is to allow values over 100% to be entered by the user while applying the policy

### Issues Resolved

https://flexera.atlassian.net/browse/FOPTS-717

https://app.flexera.com/orgs/1105/automation/applied-policies/projects/105882?policyId=644a9a3a7b2db10001f85483

### Contribution Check List

### DIFF
--- a/cost/budget_report_alerts/CHANGELOG.md
+++ b/cost/budget_report_alerts/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v2.1
 
-- Removing the 100% limit for the threshold parameter
+- Removed the 100% limit for the threshold parameter
 
 ## v2.0
 
--- a/cost/budget_report_alerts/CHANGELOG.md
+++ b/cost/budget_report_alerts/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.1
+
+- Removing the 100% limit for the threshold parameter
+
 ## v2.0
 
 - initial release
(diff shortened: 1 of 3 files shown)


degarajesh (dismissed):
LGTM

juanestrahcl (dismissed):

cost/budget_report_alerts/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.1
+
+- Removing the 100% limit for the threshold parameter
```
XOmniverse, line 5: This should be in past tense, since it is describing what was done, not what we are doing.

XOmniverse (changes requested):
I've submitted a request to alter the CHANGELOG

Also, have we confirmed that the policy still works as expected when setting values above 100%? For example, verifying that the policy still runs, that the incident output/graphs/etc. aren't broken, etc.

cost/budget_report_alerts/CHANGELOG.md
night-codes: Fixed

TDubovchenko (approved):

XOmniverse (approved):
Approved.

PR #1260, opened 2023-04-27
FOPTS-714 Add instance types to EC2 reference file

night-codes:
### Description

Support for M6a was called out by generali as being missing from our rightsizing recommendations.  See attached list of all instance types in use by Generali, we ideally need to be able to provide alternatives for all of these instance types.

Details of instance types are here: https://aws.amazon.com/ec2/instance-types/

### Issues Resolved

https://flexera.atlassian.net/browse/FOPTS-714

### DIFF
--- a/data/aws/instance_types.json
+++ b/data/aws/instance_types.json
@@ -586,6 +586,105 @@
     "enhanced_networking": true,
     "vcpu": "72"
   },
+  "c6a.large": {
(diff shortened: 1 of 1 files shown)


juanestrahcl (approved):

XOmniverse (approved):
LGTM

PR #1266, opened 2023-05-05
POL-823 OCI Bill Ingestion Fixes/Improvements

XOmniverse:
### Description

This is a minor overhaul of this policy to fix bugs, improve usability, and ensure the policy follows best practices.

### Issues Resolved

- Fixed bug that occasionally caused duplicate or missing cost data. This was caused by relying on timeModified instead of timeCreated when determining which objects to send into CBI.

- Added option to run for previous month along with relevant parameter. The README was also updated to recommend that users apply the policy twice, once for the current month and once for the previous month, to ensure that any changes made to the bill after the month ends (a common thing for cloud providers) are incorporated into the platform.

- Added regex to ensure that user-specified billing periods are valid. Self-explanatory why this would be a good thing.

- General code cleanup and modernization. In particular, some code was refactored to make more effective use of underscore.js, datasources were paired with their respective scripts instead of having a distinct section in the policy for each, and comments were added to explain the branching logic of the policy.

We do not have a shared test environment where I can link to an applied policy, but I can confirm that I have tested this new policy with a customer (Banco de Chile) and it works as expected.

### Contribution Check List

### DIFF
--- a/cost/oracle/oracle_cbi/CHANGELOG.md
+++ b/cost/oracle/oracle_cbi/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v3.0
+
+- Fixed bug that occasionally caused duplicate or missing cost data
+- Added option to run for previous month along with relevant parameter
+- Added regex to ensure that user-specified billing periods are valid
+- General code cleanup and modernization
+
 ## v2.0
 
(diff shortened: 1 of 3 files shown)


nia-vf1 (approved):
LGTM

PR #1271, opened 2023-05-10
mysql_flexible_tier_types.json Updates/Fixes

XOmniverse:
### Description

mysql_flexible_tier_types.json has been updated with the following fixes and improvements:

- up_sku and down_sku keys have been added, since not every sku cleanly upscales or downscales to something with the same name.
- Lists were resorted to be more logical and readable.
- Standard_E80ids_v4 is no longer listed as a recommended SKU size for any of the other SKUs, since it is an odd one-off SKU. There are upsize/downsize recommendations for this SKU though.


nia-vf1 (approved):
LGTM

PR #1298, opened 2023-05-29
FOPTS-710 Excessive escalation triggering

vadgun:
### Description

Both Policies are updated and using  hash_exclude, to avoid excessive triggering 

# Issues Resolved

[Jira Ticket FOPTS-710](https://flexera.atlassian.net/browse/FOPTS-710)

### Contribution Check List

### DIFF
--- a/cost/azure/idle_compute_instances/CHANGELOG.md
+++ b/cost/azure/idle_compute_instances/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v4.6
 
-- Added hash_exclude to ignore changes to the averageCPU, averages and savings, since those will change constantly
+- New escalations are no longer created if Estimated Monthly Savings or CPU Average have changed, but nothing else has.
 
 ## v4.5
 
--- a/cost/azure/idle_compute_instances/CHANGELOG.md
+++ b/cost/azure/idle_compute_instances/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v4.6
 
-- Added hash_exclude at policy definition segment
+- Added hash_exclude to ignore changes to the averageCPU, averages and savings, since those will change constantly
 
 ## v4.5
 
--- a/cost/azure/idle_compute_instances/CHANGELOG.md
+++ b/cost/azure/idle_compute_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.5
+
+- Added hash_exclude at policy definition segment
+
 ## v4.4
 
 - Fixed error that prevented Cloud Workflow actions from completing as expected.
--- a/cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
+++ b/cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
@@ -5,9 +5,9 @@ short_description "Checks for instances that are idle for the last 30 days and t
 long_description ""
 category "Cost"
 severity "low"
-default_frequency "daily"
+default_frequency "15 minutes"
 info(
-  version: "4.4",
+  version: "4.5",
   provider: "Azure",
   service: "Compute",
   policy_set: "Idle Compute Instances"
@@ -375,6 +375,9 @@ parameters "param_cpu_average_percentage","ds_azure_instance_performance"
 result "results"
 code <<-EOS
   var results = []
+  function roundNumber(num, dec) {
+    return Math.round(num * Math.pow(10, dec)) / Math.pow(10, dec);
+  }
   _.each(ds_azure_instance_performance, function(result){
     var averages = _.pluck(result.averages,"average")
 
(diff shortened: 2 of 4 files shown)


cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
```diff
@@ -376,6 +376,9 @@ parameters "param_cpu_average_percentage","ds_azure_instance_performance"
 result "results"
 code <<-EOS
   var results = []
+  function roundNumber(num, dec) {
```
XOmniverse, line 379: This isn't necessary. You can use the built-in toFixed() method to round a number, and you can use parseFloat() to convert it from a string back to a float if necessary.

Here's an example from the node console:

> number = 1.23456789
1.23456789
> number.toFixed(2)
'1.23'
> parseFloat(number.toFixed(2))
1.23

cost/azure/idle_compute_instances/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.6
+
+- Added hash_exclude at policy definition segment
```
XOmniverse, line 5: This should be more human readable and describe what the change actually does to how the policy operates rather than just literally describing what changes were made to the code.

cost/azure/unused_ip_addresses/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.6
+
+- Added hash_exclude at policy definition segment
```
XOmniverse, line 5: This should be more human readable and describe what the change actually does to how the policy operates rather than just literally describing what changes were made to the code.

XOmniverse (changes requested):

cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
```diff
@@ -387,7 +391,7 @@ code <<-EOS
       },
     0)
 
-    var final_average = parseFloat(sum/averages.length)
+    var final_average = parseFloat(sum/averages.length).toFixed(2)
```
XOmniverse, line 394: .toFixed(2) will convert the number to a string, which will adversely affect sorting. Instead, encapsulate the toFixed inside the parseFloat, like so:

var final_average = parseFloat((sum / averages.length).toFixed(2))

cost/azure/idle_compute_instances/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.6
+
+- Added hash_exclude to ignore changes to the averageCPU, averages and savings, since those will change constantly
```
XOmniverse, line 5: This still refers to much to the actual code. Keep in mind that the CHANGELOG needs to be understandable to someone who is using this policy but never looks at the actual policy template code. For example, we shouldn't mention hash_exclude at all, or refer to entries in the export statement by their code labels, since an end user would never see this stuff.

I also don't see a field called "averages" in the export statement, so I am not sure why that is mentioned or included in the hash_exclude.

An example of a better CHANGELOG entry for this:

 - New escalations are no longer created if Estimated Monthly Savings or CPU Average have changed, but nothing else has.

cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
```diff
@@ -488,20 +492,32 @@ script "js_instance_cost_mapping", type:"javascript" do
...
+
 ###############################################################################
 # Policy
 ###############################################################################
 
 policy "azure_resource_policy" do
-  validate $ds_instance_cost_mapping do
+  validate $ds_only_instances do
     summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): Azure idle compute instances found"
+    hash_exclude "averageCPU", "averages", "savings"
```
XOmniverse, line 513: What is "averages" here referring to? This isn't an entry in the export statement.

XOmniverse (changes requested):

cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
vadgun: Done, Thank you

cost/azure/idle_compute_instances/CHANGELOG.md
vadgun: Thank you, changed

cost/azure/idle_compute_instances/azure_idle_compute_instances.pt
vadgun: In tests, adding it to the hash_exclude, the results have been successful, when I remove it, the incidents continue to be generated.

Is a field contained in the object but not exported
image

XOmniverse (approved):
LGTM

PR #1301, opened 2023-06-01
Oracle Bill Ingestion README Update

XOmniverse:
### Description

Updated the README for the Oracle Bill Ingestion policy to place special emphasis on the policy frequency. I have run into multiple instances recently of users setting the frequency to "daily" instead of leaving it at the default of "15 minutes", and this causes the policy to never manage to upload all of the files and commit the bill upload.


taterdog590 (approved):
LGTM

PR #1367, opened 2023-07-21
Meta Generator Fixes

XOmniverse:
### Description

Fixes some issues with the meta policy generator.

### Issues Resolved

- It can now handle policies with resource_level set to false
- Better regex for finding parameter blocks so that the word "end" within a field doesn't break it


nia-vf1 (approved):
LGTM

PR #1408, opened 2023-08-21
POL-903 Fixed Cost Data Issue with New Azure APIs

XOmniverse:
### Description

The casing and naming conventions of some dimensions, such as the Service field, are different between the old method of ingesting Azure bills and the new method. For example, “Microsoft.Compute” is now “microsoft.compute”. This update enables our cost optimization policies to pull this data regardless of casing.

https://flexera.atlassian.net/browse/POL-903

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=64e35fc94b956f0001f2864f
https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=64e3694f2d830e000134c702

(Note: Since this is a test environment, the savings in the incident will be $0. This is just to demonstrate that the updated Optima API call itself works and does not result in any errors. I have also confirmed that this change works as expected in a client org that I was developing for, which is where I discovered the issue to begin with.)

### Contribution Check List

- (No README changes since this is just a bug fix. The policy still works the same from a user POV)

### DIFF
--- a/cost/azure/idle_compute_instances/CHANGELOG.md
+++ b/cost/azure/idle_compute_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.1
+
+- Corrected issue with policy not retrieving cost data on orgs using newer Azure bill connections
+
 ## v5.0
 
 - Renamed Subscription List parameter for consistency and accuracy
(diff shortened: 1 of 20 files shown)


nia-vf1 (approved):
LGTM

PR #1413, opened 2023-08-22
fix: fpt install from source

bryankaraffa:
### Description

Fixes issue seen here related to fpt binary not being available.  Given the rightscale.com URL, I suspect this is a deprecation and so this proposed fix replaces that step with steps that build fpt binary from source and put it in the same location toplevel of the repo

### Issues Resolved

https://github.com/flexera-public/policy_templates/actions/runs/5942493295/job/16115579747#step:3:8


XOmniverse (approved):
LGTM

PR #1467, opened 2023-09-18
Fix Azure VM Pricing JSON

XOmniverse:
### Description

For some reason, the script that ran to update this file only included VMs in the Standard tier. I reran the script and it gave the complete list of prices as expected, and this is to merge those back into the file.


nia-vf1 (approved):
LGTM

PR #1477, opened 2023-09-19
POL-921 Add Region Filtering to Azure Rightsizing Compute

XOmniverse:
### Description

This non-breaking update adds the ability to filter by region, similar to other Azure usage reduction policies. 

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=6509ac05656c530001e1762e
(Parameter is set to only allow instances in the eastus region, and this is reflected in the incident)

### Contribution Check List

### DIFF
--- a/cost/azure/rightsize_compute_instances/CHANGELOG.md
+++ b/cost/azure/rightsize_compute_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v3.3
+
+- Added ability to filter resources by region
+
 ## v3.2
 
 - Corrected issue with policy not retrieving cost data on orgs using newer Azure bill connections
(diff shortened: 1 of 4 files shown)


cost/azure/rightsize_compute_instances/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v3.3
+
+- Added ability to filter resources by region
```
nia-vf1, line 5: Is adding new parameters a breaking change and therefore this should be version 4.0?

cc - @bryankaraffa

XOmniverse: @nia-vf1 Only if they don't have default values, and the default values don't enable it to function as it did before the parameter existed.

In this case, simply not specifying a value for the new parameters will result in the policy functioning identically to the previous version, so the change is not breaking.

nia-vf1: @XOmniverse - You mean only if they do have default values?

nia-vf1 (approved):
LGTM

PR #1481, opened 2023-09-20
POL-922 AWS Superseded EC2 Instances

XOmniverse:
### Description

This is a new AWS-specific version of the Superseded Instances policy. This policy requires an AWS credential with permissions similar to the AWS Rightsizing EC2 Instances policy, but offers several advantages over the cloud-agnostic version:

- Completely updated/rebuilt code based on the updated versions of other usage reduction policies with similar functionality where appropriate
- Since instance metadata is mostly pulled from AWS itself, any instances that are actioned should no longer appear in future incidents since their instance type will no longer be superseded and the policy is gathering this metadata in near real time from the AWS API directly
- Savings information is included based on AWS list prices for instance types with currency conversion (similar to AWS Unused IP policy)
- Policy metadata and incident output have been normalized for scraping into the Optimization dashboard
- Added ability to action on recommendations directly from the incident
- Meta policy support

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=651577db656c530001e19023
(CWF action has been tested and works. The list price/savings calculation also works, but there is no account that we can test in to provide this information because the instances were created today. I've confirmed elsewhere that it does work as expected though.)

### Contribution Check List

### DIFF
--- a/cost/aws/superseded_instances/CHANGELOG.md
+++ b/cost/aws/superseded_instances/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## v1.0
+
+- initial release
--- a/cost/aws/superseded_instances/README.md
+++ b/cost/aws/superseded_instances/README.md
@@ -12,7 +12,7 @@ This policy checks all the EC2 instances in an AWS Account to determine if the i
 
 ### Policy savings details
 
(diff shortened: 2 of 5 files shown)


taterdog590 (approved):
LGTM

nia-vf1 (approved):
LGTM

PR #1463, opened 2023-09-14
fix: Refactor Azure Unused Volume and Unused IP Actions

bryankaraffa:
### Description

- Refactored the `Delete Volume` and `Create Snapshot`, and `Delete IP Address` Actions to use updated Azure APIs, improve debugging, and error handling

### Link to Example Applied Policy

https://app.flexera.com/orgs/27018/automation/applied-policies/projects/116186?policyId=650304aa3a1bb5000199025d

Tested actions with single volume, multiple volumes, and volumes that have delete locks (expect error)

### Contribution Check List

- [X] New functionality includes testing.
- [X] New functionality has been documented in the README if applicable
- [X] New functionality has been documented in CHANGELOG.MD

PR #1482, opened 2023-09-21
fix: Refactor Azure Unused Volumes, Unused IPs CWF Actions

bryankaraffa:
### Description

Refactors the CWF Actions for Azure Unused Volumes Policy and Azure Unused IP Address Policy


bryankaraffa:
These changes already merged with #1463

PR #1511, opened 2023-10-02
FOPTS-1503 Added new cost dimensions for savings calculations

JesusParada:
### Description

The following dimensions were added to the average cost per hour calculation:

- operating_system
- database_engine
- database_edition
- license_model
- deployment_option

In addition, the way of obtaining the costs per hour for each type of instance was changed. Previously, all possible combinations of instances were obtained, obtaining a cost for each combination of instance_type and region (for example, if we have 100 different instances_type and 15 regions, we would obtain 1500 average costs, even though many were not necessary according to the records obtained from the API). By adding the 5 new dimensions, this approach becomes impractical, since it would increase exponentially for each combination of operating system, database engine, database edition, etc.

So it was decided to use a map with a key composed of
` instance_type---region---operating_system---database_engine---database_edition---license_model---deployment_option`
to filter the cost records obtained from the API that match that combination, to subsequently calculate the average cost per hour of the records with purchase_option = "On demand" and thus calculate the estimated savings of the records of the other purchase_option (Reserved, Savings Plan, Spot)

### Issues Resolved

https://flexera.atlassian.net/browse/FOPTS-1503

[Applied Policy](https://app.flexera.com/orgs/1105/automation/applied-policies/projects/60073?policyId=651b7ebe66636500010089eb)

### Contribution Check List

### DIFF
--- a/cost/aws/savings_realized/aws_savings_realized.pt
+++ b/cost/aws/savings_realized/aws_savings_realized.pt
@@ -290,7 +290,6 @@ script "js_get_savings_realized", type: "javascript" do
       })
     })
   })
-  //console.log("Cost instance hours", unique_cost_inst_hrs)
 
   //apply cost_inst_hr to calculate savings realized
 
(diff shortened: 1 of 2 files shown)


cost/aws/savings_realized/aws_savings_realized.pt
```diff
@@ -257,9 +328,12 @@ script "js_get_savings_realized", type: "javascript" do
...
     //Savings Plan savings
     _.each(sp_costs, function(sp){
-      if (sp.instance_type == inst_hr.instance_type && sp.region == inst_hr.region){
+      if (sp.instance_type == inst_hr.instance_type && sp.region == inst_hr.region && sp.operating_system == inst_hr.operating_system &&
+          sp.license_model == inst_hr.license_model && sp.database_edition == inst_hr.database_edition &&
+          sp.database_engine == inst_hr.database_engine && sp.deployment_option == inst_hr.deployment_option){
+        sp_processed_counter++
         savings_realized_per_inst_hr = inst_hr.avg_od_cost_inst_hr - (sp.cost / sp.usage_amount)
-        console.log(savings_realized_per_inst_hr, sp.region, sp.instance_type)
+        //console.log(savings_realized_per_inst_hr, sp.region, sp.instance_type)
```
night-codes, line 336: Is this commented code really necessary?

JesusParada: Really  it is not, It's just for testing/debug. I'll remove all the commented console.log lines. Good catch, thanks

midir99 (approved):
LGTM

night-codes (approved):

PR #1518, opened 2023-10-04
fix: update filter to include Azure MCA

bryankaraffa:
### Description

Currently Azure Meta Parent policies will not include subscriptions under an Azure MCA due to the filter.  This updates the filter to use `substring` instead of `equal` for identifying Azure Subscriptions


XOmniverse (commented):
LGTM

XOmniverse (approved):
LGTM

PR #1520, opened 2023-10-05
POL-837 AWS Reserved Instances Recommendations Revamp

XOmniverse:
### Description

This is a revamp of the AWS Reserved Instances Recommendations policy. In addition to the usual updates, the policy also now performs currency conversion where appropriate.

Note on currency conversion:

- The AWS API that produces the recommendations also includes a currency code. Conversion is done when this currency code does not match the currency code on the Flexera org. A statement in the incident details specifies the currency converted from, the currency converted to, and the exchange rate used. This should enable users to calculate the original unmodified values if desired.

From the CHANGELOG:

- Policy now automatically converts savings from USD to local currency when appropriate
- Removed parameter to do the above manually via a user-specified exchange rate
- Added exchange rate context to incident to allow user to derive unmodified USD values when needed
- Several parameters altered to be more descriptive and intuitive to use
- `ElasticSearch` is now referred to as `OpenSearch` in keeping with current AWS naming conventions
- Normalized incident export to be consistent with other policies
- Policy now requires a valid Flexera credential

Due to the nature of the API request made in this policy, it cannot effectively be tested in our internal testing accounts. I have confirmed it works elsewhere and can provide a sample incident upon request.

### Contribution Check List


cost/aws/reserved_instances/recommendations/README.md
```diff
@@ -1,24 +1,34 @@
 # AWS Reserved Instances Recommendation
```
nia-vf1, line 1:
```suggestion
# AWS Reserved Instances Recommendations
```

nia-vf1 (approved):
LGTM

PR #1532, opened 2023-10-09
POL-844/POL-845 Azure Reserved Instances/Savings Plan Recommendations Revamp

XOmniverse:
### Description

This is a revamp of the Azure Reserved Instances Recommendations and Savings Plan Recommendations policies. In addition to the usual updates, the policy also now performs currency conversion where appropriate.

Note on currency conversion for Reserved Instances Recommendations:

- The Azure API that produces the recommendations does *not* include a currency code. The currency code is found via a call to the Azure Billing API.
- Conversion is done when the above currency code does not match the currency code on the Flexera org. A statement in the incident details specifies the currency converted from, the currency converted to, and the exchange rate used. This should enable users to calculate the original unmodified values if desired.

Note on currency conversion for Savings Plan Recommendations:

- The Azure API that produces the recommendations also includes a currency code. Conversion is done when this currency code does not match the currency code on the Flexera org. A statement in the incident details specifies the currency converted from, the currency converted to, and the exchange rate used. This should enable users to calculate the original unmodified values if desired.

Reserved Instances Recommendations CHANGELOG:

- Policy now automatically converts savings from USD to local currency when appropriate
- Added exchange rate context to incident to allow user to derive unmodified USD values when needed
- Added ability to use Subscription list parameter as either an "allow" list or a "deny" list
- Added ability to filter recommendations by region
- Several parameters altered to be more descriptive and intuitive to use
- Normalized incident export to be consistent with other policies
- Policy now requires a valid Flexera credential

Savings Plan Recommendations CHANGELOG:

- Policy now automatically converts savings from USD to local currency when appropriate
- Added exchange rate context to incident to allow user to derive unmodified USD values when needed
- Added ability to use Subscription list parameter as either an "allow" list or a "deny" list
- Several parameters altered to be more descriptive and intuitive to use
- Normalized incident export to be consistent with other policies
- Policy now requires a valid Flexera credential

Due to the nature of the API request made in these policies, they cannot effectively be tested in our internal testing accounts. I have confirmed they work elsewhere and can provide sample incidents upon request.

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #1553, opened 2023-10-17
feat: Add Policy Template Source parameter on Meta Parent Policy

bryankaraffa:
### Description

Adds a parameter which enables a Meta Parent Policy Template to use the uploaded Policy Template for the child instead of the Published Template.  This is primarily helpful for development, but may also be helpful for customers that use forked versions of our Child Policy Templates.

https://app.flexera.com/orgs/27018/automation/applied-policies/projects/116186?policyId=652ed5b4e947000001d933b3
Deployed using uploaded policy template


bryankaraffa:
@XOmniverse, @nia-vf1, @flexera-public/policy-developers -- Any thoughts for how we can test this?  Cherry pick a few Meta Parent PTs from each Cloud Provider or... ?

PR #1577, opened 2023-10-27
POL-957 AWS Rightsize EC2 CloudWatch Fix

XOmniverse:
### Description

This fixes several issues with how CloudWatch data is pulled that would occasionally cause incorrect results. Specifically, the date range we pull data for is now rounded down to the nearest hour. This is to ensure that the range is neatly sliced by the period we specify so that we only get one result back from CloudWatch for each metric.

Previously, it would sometimes imperfectly slice the data, giving us two results, the first being for less than an hour of time and the second being the number we actually needed, but when this happened, the policy grabbed the first number instead. As a backup, just in case somehow the above doesn't work, the policy now grabs the last item in the list of values instead of just grabbing [0] every time.

Additionally, I spotted a couple of issues where the lookback period was not being correctly used to adjust our request to the amount of days the user specified. Those have been fixed as well.

https://app.flexera.com/orgs/27018/automation/applied-policies/projects/116186?policyId=653bb85099745e0001587d81

### Contribution Check List

### DIFF
--- a/cost/aws/rightsize_ec2_instances/CHANGELOG.md
+++ b/cost/aws/rightsize_ec2_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.3
+
+- Fixed issue with gathering and interpreting CloudWatch metrics
+
 ## v4.2
 
 - Updated description of `Account Number` parameter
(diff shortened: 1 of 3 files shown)


nia-vf1 (approved):
LGTM

PR #1583, opened 2023-10-30
POL-853 - Azure Databricks Rightsize Compute Instances

bryankaraffa:
### Description

This policy checks all the instances associated with Azure Databricks workspaces in Azure Subscriptions for the average or maximum CPU and/or memory usage over a user-specified number of days. If the usage is less than the user provided Idle Instance CPU and/or memory percentage threshold then the Virtual Machine is recommended for deletion. 

### Issues Resolved

https://flexera.atlassian.net/browse/POL-853

https://app.flexera.com/orgs/27018/automation/applied-policies/projects/116186?policyId=653fe3d8e947000001d93586

### Contribution Check List

### DIFF
--- a/cost/azure/databricks/rightsize_compute/README.md
+++ b/cost/azure/databricks/rightsize_compute/README.md
@@ -67,11 +67,11 @@ This is the recommended method and enables a single Azure Service Principal to t
 
    You can get this from the Azure Portal or via Flexera } Automation } Credentials and get the Client ID for the Azure RM Credential that is being used.  The Azure SP that is used for other Flexera Azure Policy Templates can be used for the Databricks Policy Templates.
 
-2. Add Service Principal to all DB Workspace using the Client ID
+1. Add Service Principal to all DB Workspace using the Client ID
 
    Databricks Workspace } Admin Settings } Service Principal (i.e. `https://{workspaceUrl}/?#setting/accounts/servicePrincipals` )
 
-3. Grant Service Principal Permissions in DB Workspace
+1. Grant Service Principal Permissions in DB Workspace
 
    We currently recommend adding Service Principal to `admin` group so it can see all clusters and compute resources within the cluster.
 
--- a/cost/azure/databricks/rightsize_compute/README.md
+++ b/cost/azure/databricks/rightsize_compute/README.md
@@ -20,7 +20,7 @@ Because the virtual machine resources can often be short-lived and ephemeral, we
 
 If the lookback period is 30 days (default value), the estimated savings for a Delete is 100% the actual cost during that 30 day period for a Downsize is 50% the actual cost during that 30 day period.
 
-If the lookback period is < 30 days (1 month), we calculate what % of 30 days the lookback period is and then use that ratio to estimate the savings for a full month period (30 days).  If the actual cost during a 15day lookback period was $100, the estimated savings calculation for a Delete is `(30/15) * $100`.  In this example, a Delete would estimate $200 Potential *Monthly* Savings and a Downsize would estimate $100 Potential *Monthly* Savings.
+If the lookback period is { 30 days (1 month), we calculate what % of 30 days the lookback period is and then use that ratio to estimate the savings for a full month period (30 days).  If the actual cost during a 15day lookback period was $100, the estimated savings calculation for a Delete is `(30/15) * $100`.  In this example, a Delete would estimate $200 Potential *Monthly* Savings and a Downsize would estimate $100 Potential *Monthly* Savings.
 
 The savings is displayed in the Estimated Monthly Savings column. The incident message detail includes the sum of each resource *Estimated Monthly Savings* as *Potential Monthly Savings*.
 
@@ -30,12 +30,12 @@ The savings is displayed in the Estimated Monthly Savings column. The incident m
 
 This is required for the policy templates to be able to map each Azure Virtual Machine to a Databricks Cluster.
 
-Navigate to *Administration > Custom Tags* in Flexera and Create a new Tag Dimension
+Navigate to *Administration } Custom Tags* in Flexera and Create a new Tag Dimension
 
 | Tag Display Name | Tag Keys | Tag ID (if creating via API instead of UI) |
 | ---------------- | -------- | --- |
 | `Azure Databricks ClusterId` | `ClusterId` | `tag_azure_databricks_clusterid` |
-> *Note: These values are case-sensitive*
+} *Note: These values are case-sensitive*
 
 ### Credential configuration
 
@@ -65,31 +65,32 @@ This is the recommended method and enables a single Azure Service Principal to t
 
 1. Get Service Principal's Client ID
 
-   You can get this from the Azure Portal or via Flexera > Automation > Credentials and get the Client ID for the Azure RM Credential that is being used.  The Azure SP that is used for other Flexera Azure Policy Templates can be used for the Databricks Policy Templates.
+   You can get this from the Azure Portal or via Flexera } Automation } Credentials and get the Client ID for the Azure RM Credential that is being used.  The Azure SP that is used for other Flexera Azure Policy Templates can be used for the Databricks Policy Templates.
 
 2. Add Service Principal to all DB Workspace using the Client ID
 
-   Databricks Workspace > Admin Settings > Service Principal (i.e. `https://<workspaceUrl>/?#setting/accounts/servicePrincipals` )
+   Databricks Workspace } Admin Settings } Service Principal (i.e. `https://{workspaceUrl}/?#setting/accounts/servicePrincipals` )
 
 3. Grant Service Principal Permissions in DB Workspace
 
    We currently recommend adding Service Principal to `admin` group so it can see all clusters and compute resources within the cluster.
 
-   Databricks Workspace > Admin Settings > Groups (i.e. `https://<workspaceUrl>/?#setting/accounts/groups` )
+   Databricks Workspace } Admin Settings } Groups (i.e. `https://{workspaceUrl}/?#setting/accounts/groups` )
 
 ##### Create `OAuth2` Credential in Flexera
 
 Replace these:
 
- - `<access_token>`
- - `<flexeraProjectId>`
- - `<Credential Name>`
- - `<credentialId>`
- - `<clientId>` and `<clientSecret>`
- - `<tenantId>`
+- `{access_token}`
+- `{flexeraProjectId}`
+- `{Credential Name}`
+- `{credentialId}`
+- `{clientId}`
+- `{clientSecret}`
+- `{tenantId}`
 
 ```sh
-curl 'https://api.flexera.com/cred/v2/projects/<flexeraProjectId>/credentials/oauth2/<credentialId>' -X PUT -H 'Content-Type: application/json' -H 'Api-Version: 1.0' -H 'Authorization: Bearer <access_token>' --data-raw '{"description":"","name":"<Credential Name>","grantType":"client_credentials","tags":[{"key":"provider","value":"databricks"}],"clientCredentialsParams":{"additionalParams":{"resource":"2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"},"clientId":"<clientId>","clientSecret":"<clientSecret>"},"tokenUrl":"https://login.windows.net/<tenantId>/oauth2/token"}'
+curl 'https://api.flexera.com/cred/v2/projects/{flexeraProjectId}/credentials/oauth2/{credentialId}' -X PUT -H 'Content-Type: application/json' -H 'Api-Version: 1.0' -H 'Authorization: Bearer {access_token}' --data-raw '{"description":"","name":"{Credential Name}","grantType":"client_credentials","tags":[{"key":"provider","value":"databricks"}],"clientCredentialsParams":{"additionalParams":{"resource":"2ff814a6-3304-4ab8-85cb-cd0e6f879c1d"},"clientId":"{clientId}","clientSecret":"{clientSecret}"},"tokenUrl":"https://login.windows.net/{tenantId}/oauth2/token"}'
 ```
 
 #### Option 2: Auth to Databricks using Databricks Personal Access Token
@@ -98,14 +99,14 @@ curl 'https://api.flexera.com/cred/v2/projects/<flexeraProjectId>/credentials/oa
 
 Replace these:
 
- - `<access_token>`
- - `<flexeraProjectId>`
- - `<Credential Name>`
- - `<credentialId>`
- - `<personalAccessToken>`
+- `{access_token}`
+- `{flexeraProjectId}`
+- `{Credential Name}`
+- `{credentialId}`
+- `{personalAccessToken}`
 
 ```sh
-curl 'https://api.flexera.com/cred/v2/projects/<flexeraProjectId>/credentials/oauth2/<credentialId>' -X PUT -H 'Content-Type: application/json' -H 'Api-Version: 1.0' -H 'Authorization: Bearer <access_token>' --data-raw '{"description":"","name":"<Credential Name>","field":"Authorization","location":"header","type":"Bearer","tags":[{"key":"provider","value":"databricks"}],"key":"<personalAccessToken>"}'
+curl 'https://api.flexera.com/cred/v2/projects/{flexeraProjectId}/credentials/oauth2/{credentialId}' -X PUT -H 'Content-Type: application/json' -H 'Api-Version: 1.0' -H 'Authorization: Bearer {access_token}' --data-raw '{"description":"","name":"{Credential Name}","field":"Authorization","location":"header","type":"Bearer","tags":[{"key":"provider","value":"databricks"}],"key":"{personalAccessToken}"}'
 ```
 
 ***Important*** - When creating an Applied Policy using a Databricks Personal Access Token credential, you must provide just 1 workspace identifier (Name or ID) in `param_databricks_workspace_list`.  Personal Access Tokens are workspace-scoped and a single applied policy cannot traverse multiple Databricks workspaces with a single Personal Access Token.
--- a/cost/azure/databricks/rightsize_compute/README.md
+++ b/cost/azure/databricks/rightsize_compute/README.md
@@ -20,7 +20,7 @@ Because the virtual machine resources can often be short-lived and ephemeral, we
 
 If the lookback period is 30 days (default value), the estimated savings for a Delete is 100% the actual cost during that 30 day period for a Downsize is 50% the actual cost during that 30 day period.
 
-If the lookback period is { 30 days (1 month), we calculate what % of 30 days the lookback period is and then use that ratio to estimate the savings for a full month period (30 days).  If the actual cost during a 15day lookback period was $100, the estimated savings calculation for a Delete is `(30/15) * $100`.  In this example, a Delete would estimate $200 Potential *Monthly* Savings and a Downsize would estimate $100 Potential *Monthly* Savings.
+If the lookback period is less than 30 days (1 month), we calculate what % of 30 days the lookback period is and then use that ratio to estimate the savings for a full month period (30 days).  If the actual cost during a 15-day lookback period was $100, the estimated savings calculation for a Delete is `(30/15) * $100`.  In this example, a Delete would estimate $200 Potential *Monthly* Savings and a Downsize would estimate $100 Potential *Monthly* Savings.
 
 The savings is displayed in the Estimated Monthly Savings column. The incident message detail includes the sum of each resource *Estimated Monthly Savings* as *Potential Monthly Savings*.
 
@@ -35,7 +35,8 @@ Navigate to *Administration > Custom Tags* in Flexera and Create a new Tag Dimen
 | Tag Display Name | Tag Keys | Tag ID (if creating via API instead of UI) |
 | ---------------- | -------- | --- |
 | `Azure Databricks ClusterId` | `ClusterId` | `tag_azure_databricks_clusterid` |
-} *Note: These values are case-sensitive*
+
+> *Note: These values are case-sensitive*
 
 ### Credential configuration
 
@@ -121,6 +122,6 @@ This Policy Template does not incur any cloud costs
 
 ## Other Notes
 
-### Results where `Databricks Cluster ID` == `null`
+### Results where `Databricks Cluster ID` equals `null`
 
 These are virtual machines that the Flexera platform does not have costs for yet. If it were launched recently then we won't be able to map it (using cost method) for up to 24hrs.
--- a/cost/azure/databricks/rightsize_compute/README.md
+++ b/cost/azure/databricks/rightsize_compute/README.md
@@ -1,0 +1,122 @@
+# Azure Databricks Rightsize Compute Instances
+
+## What it does
+
+This policy checks all the instances associated with Azure Databricks workspaces in Azure Subscriptions for the average or maximum CPU and/or memory usage over a user-specified number of days. If the usage is less than the user provided Idle Instance CPU and/or memory percentage threshold then the Virtual Machine is recommended for deletion. If the usage is less than the user provided Underutilized Instance CPU and/or Memory percentage threshold then the Virtual Machine is recommended for downsizing. Both sets of Virtual Machines returned from this policy are emailed to the user.
+
+## Functional Details
+
+- The policy identifies all Virtual Machine resources using the bill data from Flexera API
+- The policy uses Azure API to get all utilization metrics for those resources during the lookback period.
+- The policy identifies all instances that have CPU and/or memory utilization below the user-specified idle thresholds and provides the relevant recommendation.
+- The recommendation provided for Idle Instances is a deletion action. The cluster or workspace can likely be deleted.
+- The recommendation provided for Underutilized Instances is a downsize action. The cluster or workspace can likely be downsized.
+
+### Policy savings details
+
+The policy includes the estimated monthly savings. The estimated monthly savings is recognized if the resource is deleted or downsized. Cost data from Flexera is used to retrieve and calculate the estimated savings.
+
+Because the virtual machine resources can often be short-lived and ephemeral, we use the actual cost during the lookback period to estimate Potential Monthly Savings.
+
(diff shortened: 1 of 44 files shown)


cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+type "policy"
+short_description "Checks for instances that have inefficient utilization for a certain lookback period (in days). \n See the [README](https://github.com/flexera-public/policy_templates/tree/master/cost/azure/databricks/rightsize_compute/) and [docs.flexera.com/flexera/EN/Automation](https://docs.flexera.com/flexera/EN/Automation/AutomationGS.htm) to learn more."
+severity "low"
+category "Cost"
+default_frequency "weekly"
+info(
+  version: "0.1.0",
+  provider: "Azure",
+  service: "Databricks",
+  policy_set: "Databricks",
```
XOmniverse, line 12: Should policy set be something more specific? Just thinking ahead to a possibility where we have multiple "rightsize PaaS instances" policies or something, where there might be a need to differentiate. 

Suggestion: "Rightsize PaaS Instances" or "Rightsize Databricks Instances" so that it retains parity with similar usage reduction policies.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+###############################################################################
+
+parameter "param_email" do
+  type "list"
+  label "Email Address"
+  description "Email addresses of the recipients you wish to notify"
+  default []
+end
+
+parameter "param_subscriptions_list" do
```
XOmniverse, line 27: To keep parity with other Azure usage reduction policies, there should be a parameter to make this an allow list or a block list, as well as a similar pair of parameters for region filtering (if that makes sense in the context of databricks)

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+parameter "param_stats_underutil_threshold_mem_value" do
+  type "number"
+  category "Statistics"
+  label "Underutilized Instance Memory Threshold (%)"
+  description "The Memory threshold at which to consider an instance to be 'underutilized' and therefore be flagged for downsizing. Set to -1 to ignore memory utilization"
+  min_value -1
+  max_value 100
+  default 40
+end
+
```
XOmniverse, line 124: The following parameters shoudl be added to match the Rightsize Compute policy:

param_exclusion_tags
param_min_savings
param_subscriptions_allow_or_deny
param_regions_allow_or_deny
param_regions_list

Code for the above (and their use) can probably just be copypasted with trivial modification from the Azure Rightsize Compute policy.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+credentials "auth_flexera" do
+  schemes "oauth2"
+  label "Flexera"
+  description "Select Flexera One OAuth2 credentials"
+  tags "provider=flexera"
+end
+
+
```
XOmniverse, line 152: Extraneous line break

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+  set_page_marker do
+    body_field "offset"
+  end
+end
+
+###############################################################################
+# Datasources & Scripts
+###############################################################################
+
+# Get all subscriptions
```
XOmniverse, line 180: Comment is unrelated to what is happening here.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+  description "Select Flexera One OAuth2 credentials"
+  tags "provider=flexera"
+end
+
+
+###############################################################################
+# Pagination
+###############################################################################
+
+# Pagination support
```
XOmniverse, line 157: Superfluous comment

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+      results.push(found)
+    })
+
+  } else {
+    results = ds_subscriptions
+  }
+EOS
+end
+
+# # Get all resource groups in each subscription
```
XOmniverse, line 265: This should be deleted outright from the version of the policy we merge into the catalog

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+      field "managedResourceGroupId", jmes_path(col_item, "properties.managedResourceGroupId")
+    end
+  end
+end
+
+datasource "ds_databricks_workspaces_filtered" do
+  run_script $js_databricks_workspaces_filtered, $ds_databricks_workspaces, $param_databricks_workspace_list
+end
+
+
```
XOmniverse, line 317: Extraneous line break

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+  result do
+    encoding "json"
+    collect jmes_path(response, "clusters[*]") do
+      field "databricks_workspace", iter_item
+      field "cluster", col_item
+    end
+  end
+end
+
+# datasource "ds_databricks_sqlwarehouses" do
```
XOmniverse, line 510: Defunct code should be removed unless it's here with the intent of being used at some point in the future

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+    // Default sinceLastTick to 0 -- we can calculate this later if we have a lastTick set
+    var sinceLastTick=0
+    // Only print one out of every 10 ticks
+    if (tickCount % 10 == 0) {
+      // Log the lastTick
+      if (lastTick > 0) {
+        sinceLastTick = Date.now() - lastTick
+      }
+      // Print debug
+      console.log("sinceLastTickMs="+sinceLastTick+" "+msg)
```
XOmniverse, line 579: Debug code should be removed from the catalog version

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+      // Update lastTick
+      lastTick = Date.now()
+    }
+    // Increment tick every time
+    // Do this last so the first log printed happens on the first tick
+    tickCount = tickCount+1;
+  }
+
+  // function computeConfidenceInterval(observations, confidenceLevel, zScore) {
```
XOmniverse, line 588: Defunct code should be removed unless it's here with the intent of being used at some point in the future

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+    resource_costs = _.filter(ds_databricks_workspaces_vm_costs, function(cost) {
+      // check if names match using case-insensitive comparison
+      // bill data sometimes makes RG uppercase, whereas the api value is lowercase
+      // console.log("finding if "+cost["resourceId"].toLowerCase()+" == "+r["resourceId"].toLowerCase())
+      return cost["resourceId"].toLowerCase() == r["resourceId"].toLowerCase()
+    })
+
+    // Check if we found cost information for this resource
+    if (resource_costs.length > 0) {
+      // console.log("resource_costs="+JSON.stringify(resource_costs))
```
XOmniverse, line 721: Debug code commented out; remove?

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+    end
+  end
+end
+
+script "js_databricks_workspaces_vm_costs", type: "javascript" do
+  parameters "iter_item", "ds_top_level_bcs", "rs_org_id", "rs_optima_host", "ds_lookback_time_formats"
+  result "request"
+  code <<-EOS
+  subscription_id = iter_item["managedResourceGroupId"].split("/")[2]
+  // console.log("subscription_id: "+subscription_id)
```
XOmniverse, line 818: More debug stuff to remove.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+            "virtualmachines": result_map[subscriptionId][location]["virtualmachines"].slice(i,i+chunk_size)
+          })
+        }
+      }
+    })
+  })
+  EOS
+end
+
+# datasource "ds_filtered_workspace_virtualmachines_grouped_result" do
```
XOmniverse, line 949: Commented out stuff should be removed unless we intend to use it later

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+        is_underutil = is_underutil_cpu || is_underutil_mem
+      }
+    }
+
+    // Only bother doing anything if we're checking at least one metric
+    if (checking_cpu || checking_mem) {
+      if (is_idle) { // First check to see if the resource is below the lower "idle" thresholds
+        vm["recommendationType"] = "Delete"
+        vm["recommendationDetails"] = [
+          "Delete Azure virtual machine ", vm["resourceId"], " ",
```
XOmniverse, line 1077: Might be worth modifying this text to be specific that this is a Databricks recommendation

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+
+datasource "ds_underutilized_instances" do
+  run_script $js_underutilized_instances, $ds_idle_and_underutil_instances
+end
+
+script "js_underutilized_instances", type: "javascript" do
+  parameters "ds_idle_and_underutil_instances"
+  result "result"
+  code <<-EOS
+  if (typeof ds_idle_and_underutil_instances["underutil_list"] != "undefined" && ds_idle_and_underutil_instances["underutil_list"].length > 0) {
```
XOmniverse, line 1124: if ds_idle_and_underutil_instances["underutil_list"]  IS undefined, the second check will generate an error and cause the policy to fail because .length is not a valid method for objects of the type undefined

Can be fixed by just removing the length check entirely, since if the length is 0, "result = ds_idle_and_underutil_instances["underutil_list"]" and "result = []" will have the same result (heh) anyway

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+EOS
+end
+
+###############################################################################
+# Policy
+###############################################################################
+
+policy "pol_databricks_optimizations" do
+  validate $ds_idle_instances do
+    summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{ len data }} Databricks compute instances that are Idle"
```
XOmniverse, line 1154: Should update this to use the name of the applied policy instead, similar to Azure Rightsize Compute.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+  }
+EOS
+end
+
+###############################################################################
+# Policy
+###############################################################################
+
+policy "pol_databricks_optimizations" do
+  validate $ds_idle_instances do
```
XOmniverse, line 1153: Would this be better as a validate_each?

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1478 @@
...
+    summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{ len data }} Databricks compute instances that are Idle"
+    detail_template <<-'EOS'
+## Summary
+
+{{ rs_project_name }} (Account ID: {{ rs_project_id }}): Databricks Workspaces with compute resources that are Idle
+
+Found {{ len data }} Databricks compute resources that were *underutilized* during the last {{ parameters.param_stats_lookback }} days.  *Idle* compute resources are those which have **less than {{ parameters.param_stats_idle_threshold_cpu_value }}% {{ parameters.param_stats_threshold }} CPU utilization** within the last {{ parameters.param_stats_lookback }} days.
+    EOS
+    check logic_or($ds_parent_policy_terminated,    eq(0,1)    ) # For Dev Only.. Always create an incident
```
XOmniverse, line 1162: This functionality should be changed for the catalog version

XOmniverse (changes requested):
A lot of these notes might only be relevant once this policy hits a 1.0 release. Not sure specifically what your vision is for what is acceptable in a pre 1.0 release, so bear that in mind if it seemed like I was being too nitpicky.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: https://github.com/flexera-public/policy_templates/blob/master/cost/azure/rightsize_compute_instances/azure_compute_rightsizing.pt#L1092-L1119
I recall initially using this approach from rightsize -- but I believe it didn't actually work, and I didn't bother troubleshooting because I already wanted to use another method that did not require statically defining the fields like we do in the result/export field section below.

JS will not attempt to process subsequent conditions in that if logic once it gets the `false` on the first statement -- 
```javascript
if (false && "string".toFixed(2)) {
	console.log("This should not run, or error.");
}
console.log("it worked!")
```
^^ plug this into [JSFiddle](https://jsfiddle.net/) to validate.  "string" objects don't have a `toFixed()` function and that would typically error if it was evaluated

I agree this is likely overkill but I think at this point in development I was constantly getting hung up on those small little exceptions and started to _check all the things.. everywhere_

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Updated in 1eb5c1ee434d05a2368db673801ff4d40c7bea0c
```javascript
"Resize Databricks Cluster "+vm["databricks_cluster_id"]+" which created Azure virtual machine ", vm["resourceId"], " ",
          "in Azure Subscription ", vm["subscriptionName"], " ",
...
```

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: I was thinking no.. because this is returning a list of VM resources (so we can show the utilization for each VM) -- but the recommendation is to Delete/Downsize the Databricks Cluster (which provisioned that ephemeral VM).

Without the validate each the summary is more as an aggregate -- and using the results from the incident (or Consolidated Incident) you can Pivot Group By `Databricks Cluster ID` > Sort by `Estimated Savings` as one method of prioritization for action

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Fixed in 2855dd50a0d92a0330e8c516580faafca532adf6

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1337 @@
...
+  description "Statistic to use when determining if an instance is idle/underutilized."
+  allowed_values "Average", "Maximum", "Minimum"
+  default "Average"
+end
+
+parameter "param_stats_interval" do
+  type "string"
+  category "Statistics"
+  label "Statistic Interval"
+  description "Interval to use for the time span.  PT1M=1 minute, PT5M=5 minutes, PT15M=15 minutes, PT30M=30 minutes, PT1H=1 hour, PT6H=6 hours, PT12H=12 hours, P1D=1 day"
```
XOmniverse, line 70: Might make more sense to make the human readable versions the actual options for the parameter and just convert them to the appropriate value in the policy itself using a conversation table:

`table = {
  '1 Minute': 'PT1M',
  '5 Minutes': 'PT5M'
...`

etc. etc.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1337 @@
...
+end
+
+parameter "param_stats_lookback" do
+  type "number"
+  category "Statistics"
+  label "Statistic Lookback Period"
+  description "How many days back to look at utilization metrics for compute resources. This value cannot be set higher than 90 because Azure does not retain metrics for longer than 90 days."
+  default 30
+  min_value 1
+  max_value 30
```
XOmniverse, line 82: This should either be changed to max_value 90 or the description should say 30 days, depending on which of the two is accurate. I believe it's the former.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
XOmniverse: I wonder if we should return a list of the items we're actually making recommendations against (the clusters) instead. I can see downsides to that too when it comes to transparency, since it wouldn't really be possible to include all of the context in a single entry for a database cluster for WHY we're making the recommendation.

Probably fine the way you did it. Just thinking out loud.

XOmniverse (changes requested):

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: I started doing that.. but then I was trying to figure out how to provide "proof" of that recommendation at the cluster scope by summarizing utilization of the VMs.. but I still saw cases where people would want that "most granular" level..

We could have a separate incident, which summarizes the results by Cluster ID.  Provide both scopes -- VM scope or Cluster Scope.  Only one of the two should provide the savings recommendations though (to prevent double count in Optimization Dashboard)

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Added in e318121d2caa8de7343ec38dc0b0d3a4a2ec651d

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1337 @@
...
+  label "Subscription Allowed List"
+  type "list"
+  description "Allowed Subscriptions, if empty, all subscriptions will be checked"
+  default []
+end
+
+parameter "param_databricks_workspace_list" do
+  label "Databricks Workspace Allowed List"
+  type "list"
+  description "Allowed Databricks Workspace, if empty, all workspaces will be checked"
```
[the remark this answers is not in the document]
bryankaraffa: Fixed in 2a54cc4f1ab68749c524e395ef4f5b62d0e2e5f2

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Example Summary now:
> Resize Databricks Cluster 0921-182737-cblrjym3 which created Azure virtual machine /subscriptions/0db6431e-7324-4088-aae6-772079b7e198/resourcegroups/bk-databricks-test-managed-resource-group/providers/microsoft.compute/virtualmachines/616a5a227b7b4e0c859db96b2737cd47 in Azure Subscription  (0db6431e-7324-4088-aae6-772079b7e198) from Standard_DS3_v2 to Standard_DS2_v2

cost/azure/databricks/rightsize_compute/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Policy savings details
+
+The policy includes the estimated monthly savings. The estimated monthly savings is recognized if the resource is deleted or downsized. Cost data from Flexera is used to retrieve and calculate the estimated savings.
+
+Because the virtual machine resources can often be short-lived and ephemeral, we use the actual cost during the lookback period to estimate Potential Monthly Savings.
+
+If the lookback period is 30 days (default value), the estimated savings for a Delete is 100% the actual cost during that 30 day period for a Downsize is 50% the actual cost during that 30 day period.
+
+If the lookback period is { 30 days (1 month), we calculate what % of 30 days the lookback period is and then use that ratio to estimate the savings for a full month period (30 days).  If the actual cost during a 15day lookback period was $100, the estimated savings calculation for a Delete is `(30/15) * $100`.  In this example, a Delete would estimate $200 Potential *Monthly* Savings and a Downsize would estimate $100 Potential *Monthly* Savings.
```
XOmniverse, line 23: "If the lookback period is { 30 days "

This { needs to be removed.

"15day"

Should be 15 day or 15-day

cost/azure/databricks/rightsize_compute/README.md
```diff
@@ -0,0 +1,126 @@
...
+### Create `Azure Databricks ClusterId` Tag Dimension in Flexera
+
+This is required for the policy templates to be able to map each Azure Virtual Machine to a Databricks Cluster.
+
+Navigate to *Administration } Custom Tags* in Flexera and Create a new Tag Dimension
+
+| Tag Display Name | Tag Keys | Tag ID (if creating via API instead of UI) |
+| ---------------- | -------- | --- |
+| `Azure Databricks ClusterId` | `ClusterId` | `tag_azure_databricks_clusterid` |
+} *Note: These values are case-sensitive*
```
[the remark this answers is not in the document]
XOmniverse: Unresolved because the extraneous } is still present. A - should be used if we're trying to signify a bulleted item.

cost/azure/databricks/rightsize_compute/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+- Azure
+
+## Cost
+
+This Policy Template does not incur any cloud costs
+
+## Other Notes
+
+### Results where `Databricks Cluster ID` == `null`
```
XOmniverse, line 124: Might be worth rephrasing this in a way that would be less obtuse to non-programmers. Normies don't know what == means.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1360 @@
...
+  description "Statistic to use when determining if an instance is idle/underutilized."
+  allowed_values "Average", "Maximum", "Minimum"
+  default "Average"
+end
+
+parameter "param_stats_interval" do
+  type "string"
+  category "Statistics"
+  label "Statistic Interval"
+  description "Interval to use for the time span."
```
XOmniverse, line 70: Might need more detailed description, either here or in the README.

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
```diff
@@ -0,0 +1,1360 @@
...
+      results.push(found)
+    })
+
+  } else {
+    results = ds_subscriptions
+  }
+EOS
+end
+
+#get filtered resource groups for databricks
```
XOmniverse, line 274: Is this leftover from a datasource the policy no longer contains?

XOmniverse (changes requested):

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Removed in 851728c20c2a7cc5ca94dac013c36e98a3043cad

cost/azure/databricks/rightsize_compute/README.md
bryankaraffa: Changed `==` to `equals` in 851728c20c2a7cc5ca94dac013c36e98a3043cad

cost/azure/databricks/rightsize_compute/README.md
bryankaraffa: Changes to `15-day` and `{` to `less than` in 851728c20c2a7cc5ca94dac013c36e98a3043cad

cost/azure/databricks/rightsize_compute/README.md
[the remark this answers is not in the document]
bryankaraffa: `}` changed to `>` in 851728c20c2a7cc5ca94dac013c36e98a3043cad

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Attempted to improve this in 39373a1e91af9929bea5ec0867178baf85f04ee1

@XOmniverse -- let me know if this helps provide context needed

cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt
bryankaraffa: Added in most recent commit

XOmniverse (approved):
LGTM

PR #1598, opened 2023-11-01
fix: name `AWS Ensure IAM Users Receive Permissions Only Through Groups`

bryankaraffa:
### Description

We have 2 PTs with same name.. that should never happen.  Looks like this one was incorrectly named


XOmniverse (approved):
LGTM

PR #1610, opened 2023-11-06
FOPTS-2380 Modify S3 Object Policy template to prevent tag calls if user don't input tags

vadgun:
### Description

Modify temporally a bucket list to prevent timeout from large buckets, and revert that changes once successfully results

Additionally, if user did specify exclude tags, retrieve tags only for objects we will be raising incident for, instead of making calls for all discovered objects

### Issues Resolved

[FOPTS-2380](https://flexera.atlassian.net/browse/FOPTS-2380)

[Applied policy with all buckets](https://app.flexeratest.com/orgs/1105/automation/applied-policies/projects/60073?policyId=654b49063411070001af400a)

### Contribution Check List

### DIFF
--- a/cost/aws/object_storage_optimization/CHANGELOG.md
+++ b/cost/aws/object_storage_optimization/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## v3.2
 
-- Added a filter to the S3 Objects that will be moved to deeper or glacier storage class.
+- Optimized policy code to speed up execution.
+- Tags for objects are not requested if the parameter Tag exclude is empty.
 - Added a new parameter `Bucket List`.
 
 ## v3.1
--- a/cost/aws/object_storage_optimization/CHANGELOG.md
+++ b/cost/aws/object_storage_optimization/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## v3.2
 
-- Added a filter to the s3objects that will be moved to deeper or glacier storage class
-- Added a new parameter `Bucket list`
+- Added a filter to the S3 Objects that will be moved to deeper or glacier storage class.
+- Added a new parameter `Bucket list`.
 
 ## v3.1
 
--- a/cost/aws/object_storage_optimization/CHANGELOG.md
+++ b/cost/aws/object_storage_optimization/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v3.2
 
 - Added a filter to the s3objects that will be moved to deeper or glacier storage class
+- Added a new parameter `Bucket list`
 
 ## v3.1
 
--- a/cost/aws/object_storage_optimization/CHANGELOG.md
+++ b/cost/aws/object_storage_optimization/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v3.2
+
+- Added a filter to the s3objects that will be moved to deeper or glacier storage class
+
 ## v3.1
 
 - Updated description of `Account Number` parameter
--- a/cost/aws/object_storage_optimization/CHANGELOG.md
+++ b/cost/aws/object_storage_optimization/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## v3.2
 
 - Optimized policy code to speed up execution.
-- Tags for objects are not requested if the parameter Tag exclude is empty.
+- Tags for objects are not requested if the parameter `Exclude Tag` is empty.
 - Added a new parameter `Bucket List`.
 
 ## v3.1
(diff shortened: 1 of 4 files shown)


cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
```diff
@@ -149,7 +155,62 @@ datasource "ds_aws_buckets_with_region" do
...
-  run_script $js_parse_buckets, $ds_aws_buckets_with_region
+  run_script $js_parse_buckets, $ds_aws_buckets_with_region, $param_bucket_list
+end
+
+script "js_parse_buckets", type: "javascript" do
+  parameters "buckets", "param_bucket_list"
+  result "results"
+  code <<-EOS
+    results = []
+    if (param_bucket_list.length > 0){
```
XOmniverse, line 166: There's a lot of duplicate code here between the "if" section and the "else" section. Can this be consolidated in such a way that the if-else mostly just handles the difference between the two?

One method would be to use underscore.js to filter the bucket list via _.filter (using an if-else statement to either create a new filtered list or just set the list equal to the ds_aws_buckets_with_region parameter), and then do all the other work after the if-else statement:

https://underscorejs.org/#filter

cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
```diff
@@ -178,9 +239,72 @@ datasource "ds_aws_list_s3_objects" do
...
+            "storage_class": s3Object["storage_class"],
+            "last_modified": s3Object["last_modified"],
+            "modify_storage_class_to": modify_storage_class_to,
+            "object_size": s3Object["object_size"]
+          });
+        }
+      }
+    });
+    results = _.sortBy(results, 'region');
+    console.log(results)
```
XOmniverse, line 301: console.log statements are used for debugging and should be removed from production policies

cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
```diff
@@ -205,67 +329,28 @@ datasource "ds_aws_list_s3_objects_with_tags" do
...
-    var deep_archive_date = new Date();
-    var glacier_days_to_move = parseInt(param_glacier_days,10);
-    var deep_archive_days_to_move = parseInt(param_deep_archive_days,10);
-    glacier_date = glacier_date.setDate(glacier_date.getDate() - glacier_days_to_move);
-    deep_archive_date = deep_archive_date.setDate(deep_archive_date.getDate() - deep_archive_days_to_move);
-    glacier_date = (new Date(glacier_date)).setHours(23, 59, 59, 999);
-    deep_archive_date =(new Date(deep_archive_date)).setHours(23, 59, 59, 999);
 
     _.each(ds_aws_list_s3_objects_with_tags, function(s3Object){
+      console.log(s3Object)
```
XOmniverse, line 353: console.log statements are used for debugging and should be removed from production policies

XOmniverse (changes requested):
Requested some changes. Most are fairly minor.

cost/aws/object_storage_optimization/aws_object_storage_optimization.pt
vadgun: Hi Shawn, Thank you for the review, changes are addressed.

JesusParada (dismissed):
Changes look good to me, and I think solve the ticket mentioned in the description

cost/aws/object_storage_optimization/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v3.2
+
+- Added a filter to the S3 Objects that will be moved to deeper or glacier storage class.
```
XOmniverse, line 5: I'm not clear on what this means. Could you elaborate?

XOmniverse (changes requested):

cost/aws/object_storage_optimization/CHANGELOG.md
night-codes: The description does not reflect the essence of the changes. This filter was already present, but filtering took place after receiving tags. Now we filter filter and then receive tags.

XOmniverse: So if I understand correctly, the policy's functionality did not change, but we rearchitected it a bit so that it will be more efficient. In that case, this line should be something like:

- Optimized policy code to speed up execution

The key is to remember that the CHANGELOG is going to be read by ordinary users who might have no knowledge or awareness of code, programming, etc. They need the bottom line of how the changes impact them; in this case, we're making the policy execute more quickly.

XOmniverse (changes requested):

cost/aws/object_storage_optimization/CHANGELOG.md
vadgun: Thank you Shawn, have a great weekend

night-codes (approved):
LGTM

PR #1611, opened 2023-11-07
Fix/meta parent templates email condition

bryankaraffa:
### Description

Fixes email used to filter out catalog policy template IDs


taterdog590 (approved):
LGTM

PR #1615, opened 2023-11-09
POL-967 Deprecate AWS GP3 Upgradeable Volumes Policy

XOmniverse:
### Description

This policy is no longer being updated. The [AWS Rightsize EBS Volumes](https://github.com/flexera-public/policy_templates/tree/master/cost/aws/rightsize_ebs_volumes/) policy should be used for these recommendations instead.

Note that, unlike the AWS Rightsize EBS Volumes policy, this policy reports on IO1 and IO2 volumes. These volumes are high performance volumes, so changing them to GP3 will result in a performance downgrade and may cause issues for workloads that rely on this performance.

Effectively, this policy was just recommending the above change without regard for actual usage data, which is a bit reckless and shouldn't be done. A future version of the AWS Rightsize EBS Volumes policy may include this functionality, but it would be alongside actual analysis of metrics to determine if the downgrade makes sense.

N/A. Changes have no impact on functionality or policy execution.

### Contribution Check List

### DIFF
--- a/cost/aws/gp3_volume_upgrade/CHANGELOG.md
+++ b/cost/aws/gp3_volume_upgrade/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.3
+
+- Deprecated: This policy is no longer being updated. Please see policy README for more information.
+
 ## v4.2
 
 - Updated description of `Account Number` parameter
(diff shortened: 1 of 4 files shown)


taterdog590 (approved):
LGTM

PR #1666, opened 2023-12-05
POL-971 Azure Reserved Instances Utilization - update to use Modern Azure APIs

nia-vf1:
### Description

We are currently migrating customers' bill configurations to the new Azure method, however this policy still uses legacy EA APIs, which will be deprecated as part of the migration.

This is a change to update the policy to move away from the legacy EA APIs and utilize the modern Azure APIs.

### Issues Resolved

Moving the modern Azure APIs resolves the issue of having a dependency on the Azure EA Key as an Automation Credential.

https://app.flexera.com/orgs/33693/automation/applied-policies/projects/135064?policyId=6570714a55d35d0001a9053c

### Contribution Check List

### DIFF
--- a/cost/azure/reserved_instances/utilization/CHANGELOG.md
+++ b/cost/azure/reserved_instances/utilization/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## v3.0
 
-- Updated policy to move away from Azure Legacy EA API endpoints and use the Azure Modern API endpoints due to deprecation in 2024
-- Updated credential required for policy - it now requires `azure_auth` credential due to EA Key deprecation in 2024
-- Removed `param_enrollment_id` parameter as no longer required
+- Updated policy to move away from Azure Legacy EA API endpoints and use the Azure Modern API endpoints due to EA Key deprecation in 2024
+- Updated credential required for policy - it now requires an Azure Resource Manager credential due to EA Key deprecation in 2024
+- Removed "Enrollment ID" parameter as no longer required
 - Added "Reservation Expiration Date" field to policy incident
 
 ## v2.8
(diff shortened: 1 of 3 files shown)


nia-vf1:
> One additional question (not necessarily a change request): is "isfratio.blob.core.windows.net" an official repository of Azure info? I just want to make sure we're not relying on third party data that we can't vet or control and hasn't been affirmed by Microsoft as being correct.

@XOmniverse - an official Azure thing, see https://learn.microsoft.com/en-us/azure/virtual-machines/reserved-vm-instance-size-flexibility#instance-size-flexibility-ratio-for-vms
image

cost/azure/reserved_instances/utilization/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v3.0
+
+- Updated policy to move away from Azure Legacy EA API endpoints and use the Azure Modern API endpoints due to deprecation in 2024
+- Updated credential required for policy - it now requires `azure_auth` credential due to EA Key deprecation in 2024
```
XOmniverse, line 6: Names of credentials, parameters, etc. in the CHANGELOG should match what the user would see, not how things are identified in code. For example, instead of `azure_auth`, we can say it now requires an Azure Resource Manager credential, and we should replace `param_enrollment_id` with the name of the parameter the user would see in the UI.

XOmniverse (changes requested):
Just some minor CHANGELOG updates. Ping me once that's fixed and I'll approve.

cost/azure/reserved_instances/utilization/CHANGELOG.md
nia-vf1: Resolved (I think)

PR #1678, opened 2023-12-08
POL-981 Azure Untagged Resources Revamp / Untagged Virtual Machines

XOmniverse:
### Description

This PR is for several related things:

- `/data/azure/tag-support.csv` has been replaced with `/data/azure/resource_types.json` due to JSON being natively supported by the policy engine. It also contains information unrelated to tag support, so this new naming is more accurate and will allow for the file to be extended with more resource type related metadata if ever needed. The below `Azure Untagged Resources` policy is the only policy currently making use of this file, so this change should have no impact on other policies.

- `Azure Untagged Resources` policy has been completely rebuilt from the ground up. Core functionality is the same, but new filtering features were added. From the CHANGELOG:
  - Added ability to filter resources by tag key, tag key==value, or using regex
  - Added ability to filter resources by region
  - Added ability to filter resources by Azure resource type
  - Added ability to use all filters as an allow list or a deny list
  - Policy now requires a valid Flexera One credential
 
- `Azure Untagged Virtual Machines`: This is a new policy that only checks tags for virtual machines, but due to this narrowed focus, has more functionality than the `Azure Untagged Resources` policy. The incident reports on additional VM-specific metadata, and the policy allows for powering off or deleting instances in addition to tagging them.

- The above policies now support a variety of conditionals as well as regex. Note that, while I have intentionally not documented this since it is not "to spec", using a single = will also work in order to account for user error when entering in values. From the READMEs:
  - *Tags* - The policy will report resources missing the specified tags. The following formats are supported:
    - `Key` - Find all resources missing the specified tag key.
    - `Key==Value` - Find all resources missing the specified tag key:value pair and all resources missing the specified tag key.
    - `Key!=Value` - Find all resources that have the specified tag key:value pair.
    - `Key=~/Regex/` - Find all resources where the value for the specified key does not match the specified regex string and all resources missing the specified tag key.
    - `Key!~/Regex/` - Find all resources where the value for the specified key matches the specified regex string.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=658c560e9f8a200001ecbc27
https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=658c559e7cc18300018b402d

Note: Azure Untagged Resources policy has been slightly modified from this PR for testing in order to use the `/data/azure/resource_types.json` file in this branch since it won't exist in the master branch until after this PR is merged.

### Contribution Check List


taterdog590 (approved):
LGTM

PR #1691, opened 2023-12-13
POL-1006 Google Idle Persistent Disk Recommender Permissions Update

XOmniverse:
### Description

Two permissions are required for taking actions against disks raised in this policy that are not listed in the README. This adds them to the README.


nia-vf1 (approved):
LGTM

PR #1699, opened 2023-12-18
POL-983 - Updated sql_service_tier_types instance sizes for GP5

mburmeig:
### Description

Fixes SQL mapping based on Microsoft docs.
Pepsi noticed we were making recommendations to upsize/downsize instances that did not exist in the microsoft docs.
This was causing their recommendations to be incorrect. 
https://azure.microsoft.com/en-us/pricing/details/azure-sql-managed-instance/single/

### Issues Resolved

added correct mapping for GP5

https://app.flexera.com/orgs/34088/automation/applied-policies/projects/135557?policyId=6573381d55d35d0001a91b4b

### Contribution Check List

- [NA ] New functionality includes testing.
- [NA ] New functionality has been documented in the README if applicable
- [ NA] New functionality has been documented in CHANGELOG.MD


data/azure/sql_service_tier_types.json
```diff
@@ -123,56 +123,36 @@
       "down": null
     },
     "GP_Gen5_4": {
-      "up": "6",
-      "down": "2"
-    },
-    "GP_Gen5_6": {
       "up": "8",
-      "down": "4"
+      "down": null
```
XOmniverse, line 127: Should this be null? Wouldn't this downsize to 2 e.g. GP_Gen5_2?

XOmniverse (changes requested):
Found one possible mistake.

data/azure/sql_service_tier_types.json
mburmeig: I was going by the docs noted in the ticket that does not show a 2 instance.
BK noted that we should not use the docs at the URL I posted and we should go by these docs.
https://learn.microsoft.com/en-us/azure/azure-sql/database/resource-limits-vcore-single-databases?view=azuresql

It seems the issue noted is one specific to pepsi and not an issue with the SQL mapping file.

XOmniverse: @mburmeig Then that size shouldn't be listed in the document at all.

Let me know if you've confirmed that this change is unneeded and we can delete this PR.

PR #1719, opened 2024-01-03
FOPTS-2229 Added a way to url decode the skiptoken

vadgun:
### Description

Replaced the jq function used from rt

### Issues Resolved

[FOPTS-2229](https://flexera.atlassian.net/browse/FOPTS-2229)

https://app.flexera.com/orgs/28576/automation/applied-policies/projects/126887?policyId=659c762c9f8a200001ecef8a

If you don't have access to testing org 28576, you can watch the following GIF that shows the fixed error related to pagination:

itam_report_fixed_demo

### Contribution Check List

### DIFF
--- a/operational/itam/schedule_itam_report/CHANGELOG.md
+++ b/operational/itam/schedule_itam_report/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.1.2
 
-- Added a filter to convert `%3D` to his corresponding `=` symbol to prevent bad api token request.
+- Before this change the policy wont be able to retrieve all the corresponding data because the request has an extra symbol, now a filter was applied that convert to the correct value to prevent a bad request
 
 ## v0.1.1
 
--- a/operational/itam/schedule_itam_report/CHANGELOG.md
+++ b/operational/itam/schedule_itam_report/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.1.2
 
-- Before this change the policy wont be able to retrieve all the corresponding data because the request has an extra symbol, now a filter was applied that convert to the correct value to prevent a bad request
+- Fixed pagination bug that caused the `skipToken` query parameter to be sent with extra equal sign characters in the URL, thus making an HTTP request fail.
 
 ## v0.1.1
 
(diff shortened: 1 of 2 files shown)


operational/itam/schedule_itam_report/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.1.2
+
+- Added a way to decode the pagination token
```
XOmniverse, line 5: This should be reworded in a way that would make sense to a non-technical person that wants to know what this change means for them. Explain the bug or problem they would have experienced that this fixes rather than describing it as you would to another engineer.

XOmniverse (changes requested):
I requested a change to the CHANGELOG. Also, can I get confirmation that this policy has been tested? The testing box is unchecked and there is no link to an applied policy.

operational/itam/schedule_itam_report/CHANGELOG.md
XOmniverse: The new version still has the same fundamental issue. "Added a filter to convert `%3D` to his corresponding `=` symbol to prevent bad api token request." still won't make much sense to a non-technical user.

Think of it from the user's point of view. What was the problem that they may have encountered before this change? Describe that problem in the terms that the user would understand, not an engineer, and then say you fixed that problem in the CHANGELOG.

vadgun:
Ill raise a new one after be Able to Test, it seems something about my credentials that prevent the use of Flexera One RQA environment. Thank you

midir99:
adfasdf

midir99 (approved):

operational/itam/schedule_itam_report/CHANGELOG.md
midir99: Hey @XOmniverse, @vadgun

I applied the changes to make the text more user friendly, we agreed on using the following:

Fixed a pagination bug that caused an error when the policy template requested information from the IT Asset Management (ITAM) server, thus failing the execution.

night-codes (approved):

XOmniverse (approved):
LGTM

PR #1727, opened 2024-01-08
POL-958 Add 'Minimum Age' to Azure Rightsize SQL

XOmniverse:
### Description

Added optional `Minimum Age (Days)` parameter to filter results by age. This is for users that want to avoid reporting on freshly created databases that, as a result of their newness, have not had any connections and would therefore be seen as "unused" by the policy.

This is not a breaking change since the default value of this parameter is 0 and this functions just like the policy did without the parameter.

From the README:

- *Minimum Age (Days)* - The minimum age, in days, since a SQL database was created to produce recommendations for it. Set to 0 to ignore age entirely.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=659c1da07cc18300018b73f7

(Note: The above does not produce an incident because we don't have any old SQL servers in our test account. If you review the log, you'll see that `ds_azure_sql_databases` returns two system databases, and the filter correctly filters out one of them based on age for `ds_azure_sql_databases_age_filtered`)

### Contribution Check List

### DIFF
--- a/cost/azure/rightsize_sql_instances/CHANGELOG.md
+++ b/cost/azure/rightsize_sql_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.3
+
+- Added optional `Minimum Age (Days)` parameter to filter results by age
+
 ## v4.2
 
 - Policy action error logging modernized and now works as expected in EU/APAC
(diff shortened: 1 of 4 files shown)


nia-vf1 (approved):
LGTM

PR #1929, opened 2024-03-19
POL-1152 Dangerfile More Tests/Fixes

XOmniverse:
### Description

This Dangerfile update adds the following tests:
- Raise warning if policy has publish set to false
- Raise warning if policy's name has changed
- Raise error if modified Python scripts contain errors

The following fixes were added:
- README permission tester now correctly detects if the * line about special permissions is present or not.
- Github PR now accepts FOPTS in the title in addition to POL to account for FOPTS JIRA tickets.

The following additional changes were made:
- README permission tester no longer checks for the JSON example in AWS policy READMEs.
- Github testing workflow has been modified to install Python 3 and the pylint pip package. This is for the Python linting functionality that was added. The `.pip_requirements` file has been added and contains a list of pip packages/versions to install when the workflow executes.

### Contribution Check List


bryankaraffa (commented):
I don't think we should include a JSON example yet until we are compiling that part of the READMEs.. it's just another place that needs to be manually updated (and can get out of sync)

My thinking was the list of permissions is enough.. and eventually when we switch to compiling the permission section in the READMEs we can compile the IAM Permission Policy Exmaple as well

idc too much to block a merge on this, but that was my strategy for PTs today

bryankaraffa (approved):

PR #1947, opened 2024-03-21
Rakefile Fix Pt 2

XOmniverse:
### Description

Fixed for real this time.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,7 +15,8 @@ task :generate_policy_list do
   # Preparation for getting information from Github repository
   repo_name = "flexera-public/policy_templates"
   branch = "master"
-  github_client = Octokit::Client.new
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -15,7 +15,8 @@ task :generate_policy_list do
   # Preparation for getting information from Github repository
   repo_name = "flexera-public/policy_templates"
   branch = "master"
-  github_client = Octokit::Client.new
+  github_api_token = `echo $GITHUB_API_TOKEN`.strip
```
bryankaraffa, line 18:
```suggestion
  github_api_token = ENV["GITHUB_API_TOKEN"].strip
```
Shouldn't even need the strip imo.. but I think this will be a little safer and the value potentially less likely to get exposed in logs

bryankaraffa (approved):

PR #1953, opened 2024-03-22
POL-1163 Dangerfile Improvements

XOmniverse:
### Description

Dangefile improvements:
- New tests: Raise error if JSON/YAML file is added/changed within a directory intended for storing policies
- New tests: Raise error if JSON/YAML files fail a lint test
  - New gems `jsonlint` and `yaml-lint` have been added to `Gemfile` and `Gemfile.lock` to accomodate this
- New test: Raise error if latest version number in CHANGELOG does not match policy
- New test: Raise error if path statement for datasource whose host is `raw.githubusercontent.com` and whose path contains `/policy_templates/` is not pointed to a file that exists in the repository.
  - An error will also be raised if the link does not point directly to the `flexera-public` repository, such as old rightscale links that still technically work due to forwarding but that we should stop using. 
- New test: Raise warning if any API connections (datasource, script, or Cloud Workflow) are configured to use HTTP instead of HTTPS
- New test: Raise warning if improperly formatted comma-separation is detected. Warning because false positives are likely but should be minimal
- New test: Raise error if there are orphaned code blocks e.g. code blocks that are not referenced or used elsewhere in the policy.
- Updated test: Important Files Modified test is now a single consolidated warning to save space and be more readable.

Dangefile fixes:
- Fix: README permission test no longer incorrectly expects presence of cloud-specific credentials simply because the name of the policy references that cloud. We have a handful of cloud-specific policies that only need a Flexera credential.
  - A new policy test has been added to compare the credentials in the policy itself to the ones in the README instead.
- Fix: CHANGELOG formatting test no longer incorrectly reports indented secondary lists as a problem.
- Fix: Warning about missing READY-TO-REVIEW label changed from warning to message, since adding this label does not retrigger Dangerfile and remove the warning.

Rakefile improvements:
- Rakefile now obtains Github credential via the ENV keyword rather than a bash echo command. This should be more secure.

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #1963, opened 2024-03-27
POL-1180 Dangerfile Indentation Test Fix

XOmniverse:
### Description

Modifies Dangerfile indentation test for policies to only check indentation for EOS/EOF if it's for JavaScript code to avoid issues with lack of indentation due to a need to properly render markup/markdown languages for other fields.

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #1971, opened 2024-03-28
POL-1185 Dangerfile Outdated Terminology Update

XOmniverse:
### Description

Updates Outdated Terminology test for Dangerfile to look only for instances of "rightscale" and "optima" surrounded by whitespace so that false positives are not raised for built-in policy constructs like `rs_optima_host` or for API endpoints that are still in active use.

Also updates test to ignore contents of data/ and tools/ since these are autogenerated assets. Issues should be corrected in the actual source files, not in generated assets and tools.


nia-vf1 (approved):
LGTM

PR #2014, opened 2024-04-02
POL-1189 Change History Schedule Fix

XOmniverse:
### Description

Triggering the Change History workflow with every PR is causing externalities for other functionality by exhausting the rate limit of the Github API. To fix this, the workflow is now configured to run weekly instead of every time there is a pull request. The workflow can also be triggered manually if a mid-week update is ever needed for some reason.

All workflows set to run on a weekly basis have also been modified so they all execute on different hours on Sunday to avoid any possibility of overloading something by running several jobs all at the same time.


bryankaraffa (approved):
Discussed a few other options and this seems like a quick resolution for now

PR #2079, opened 2024-04-17
POL-1202 Currency Conversion Fix

XOmniverse:
### Description

- New Functionality: Policy can now automatically set the organization's currency to match the `Currency To` parameter. This setting defaults to being disabled to maintain parity with previous versions.
- Fixes issue where policy would fail if the org had adjustment rules set for future months. Payload is now intentionally sorted by month first to ensure that this doesn't cause issues.
- Removed the Email Addresses parameter because it's not actually used by the policy. Updated the major version number since this is technically a breaking change.
- Made the JSON stringified version of the payload for the adjustments API a separate datasource to assist in debugging when errors occur, but this has no impact on functionality from a user standpoint.
- Various tweaks were made to pass linting tests.

Tested in client environment where issue was reported.

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #2097, opened 2024-04-19
POL-1208 AWS Unused IP Address Cost Fix

XOmniverse:
### Description

It was discovered that the AWS Unused IP Address policy now sometimes fails on execution. Further digging revealed that the AWS Price List API no longer provides pricing for unattached IP addresses. This is likely because AWS now charges for all IP addresses, not just unattached ones. I also discovered the following:

- Pricing for attached IPs is consistent regardless of region in the results; all IPv4 addresses cost $0.005 USD/hour.
- The Price List API does not bother to return pricing results for a majority of AWS regions.

The short version is that the AWS Price List API doesn't really bother to provide granular IP address pricing information anymore. Because of this, and because AWS's own docs specifically outline a single consistent price, the price is now hardcoded into the policy rather than retrieved from an API.

This does mean that, if the price model changes, the policy itself will need to be updated, but that would be true regardless, since that would likely also mean that the Price List API's output would change as well. This seems to be the least bad solution until and unless AWS provides a means via API to get region-specific pricing for unattached IP addresses.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=66228520352c6d87fe314155

### Contribution Check List


data/policy_permissions_list/master_policy_permissions_list.yaml
```diff
@@ -1021,7 +1021,7 @@
       required: true
 - id: "./cost/aws/unused_ip_addresses/aws_unused_ip_addresses.pt"
   name: AWS Unused IP Addresses
-  version: '8.1'
+  version: 9.0.0
```
nia-vf1, line 1024: I notice the version field values for other templates in this list are strings. Not sure why this one looks like a number

Can see it's the same for the Policy Template Synchronization version which follows the same notation `2.0.0`

nia-vf1 (approved):
LGTM

PR #2137, opened 2024-04-29
POL-1217 Email Cost Optimization Recommendations: Always Send Incident

XOmniverse:
### Description

Added the following parameter to the `Email Cost Optimization Recommendations` policy:

- *Always Email Incident* - Whether or not to always email the incident even if no new items were added to the recommendations since the policy's last execution.

This works by adding a new incident field that is set to an empty string if this parameter is set to "No" and is set to the current date and time if the parameter is set to "Yes". Since the current date and time of policy execution will always be a unique value, this should have the desired effect.

Also added support for some newer recommendations policies that didn't exist when this policy was last updated.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=662fe40e352c6d87fe316180

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #2153, opened 2024-05-02
POL-1221 Usage Report Policies - Update BC Filter to support Child Billing Centers

nia-vf1:
### Description

Usage Reports for AWS and Azure (e.g., Amount of Instance Memory Used, Number of Instance vCPUs Used, etc.) currently allow the user to define a list of Billing Centers to report on which can either be allowed/filtered in or denied/filtered out.

The current logic only supports filtering on top-level billing centers, which means that child billing centers cannot be specifically reported on.

This is a change to add support for filtering on child billing centers when applying the policy to generate a report.

### Issues Resolved

Fixes issue where only top-level billing centers could be filtered on. Policy now additionally supports filtering on child billing centers.

Tested in customer tenant, and working as expected. Some examples below of chart produced in policy incident:

AWS Instance vCPUs Used (all Billing Centers i.e., no BC filter) - demonstrating previous functionality still works:
image

Azure Instance Memory Used (for one Child Billing Center) - demonstrating new functionality works:
image

### Contribution Check List

- [ ] ~~New functionality has been documented in the README if applicable~~

### DIFF
--- a/operational/aws/total_instance_hours/CHANGELOG.md
+++ b/operational/aws/total_instance_hours/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v4.1.0
+
+- Fixed issue where only top-level billing centers could be filtered on. Policy now additionally supports filtering on child billing centers.
+
 ## v4.0
 
 - Added ability to filter the report for a list of Billing Centers that can either be allowed or denied.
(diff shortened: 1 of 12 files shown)


XOmniverse (approved):
LGTM

PR #2164, opened 2024-05-03
SQ-7228 Incorrect currency code and inflated savings bug

midir99:
### Description

- Adds a new parameter that lets the user decide on the format for the currency shown in the incident's chart: code (e.g. USD) or symbol (e.g. $).
- Fixes a bug that showed inflated savings on the instances.

### Issues Resolved

https://flexera.atlassian.net/browse/SQ-7228

https://app.flexeratest.com/orgs/1105/automation/applied-policies/projects/107982?policyId=66353ae2e2d700d7f57fe119

### Contribution Check List

### DIFF
--- a/cost/azure/savings_realized/CHANGELOG.md
+++ b/cost/azure/savings_realized/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v3.9
+
+- Added parameter to select between showing the currency code or the currency symbol in the incident report
+
 ## v3.8
 
 - Added support for organizations with Microsoft MCA accounts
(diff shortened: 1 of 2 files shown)


XOmniverse (changes requested):
What is the impetus for this change? Shouldn't the policy just obtain the currency code via the API and use what the API returns?

Do we have a use case where someone has their Flexera org configured to use one currency but wants this report to be in a different currency?

TDubovchenko:
@XOmniverse this parameter is to choose between using currency code vs currency symbol (show USD vs $), not to choose a different currency. The motivation for that is that for one of the customers using DKK currency, the generated chart shows "kr" as currency symbol on Y axis. As customer sees DKK in Flexera UI, they asked for consistency. It seems that there is no API to get if code or symbol is used in UI, the additional parameter is added. 
Nicer approach would be to apply the same logic UI uses to show code vs symbol and apply it in the policy.

XOmniverse (approved):
LGTM

PR #2173, opened 2024-05-06
POL-1225 Azure Rightsize SQL Databases: New Incident Fields

XOmniverse:
### Description

New fields added to incident output: `Last Accessed` and `Status`

Note that this required modifying the policy to list SQL servers and then list SQL databases for each server instead of pulling the latter directly from a more general resource API. This means the policy may take longer to execute than it used to.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=66391ad3f5c05009ea60bcc0

(Also tested in client environment where incident output could be verified)

### Contribution Check List


cost/azure/rightsize_sql_instances/azure_rightsize_sql_instances.pt
```diff
@@ -508,14 +542,23 @@ datasource "ds_azure_sql_database_metrics" do
...
+    field "region", val(iter_item, "region")
+    field "serverId", val(iter_item, "serverId")
+    field "serverKind", val(iter_item, "serverKind")
+    field "serverName", val(iter_item, "serverName")
+    field "serverTags", val(iter_item, "serverTags")
+    field "serverFQDN", val(iter_item, "serverFQDN")
+    field "serverState", val(iter_item, "serverState")
+    field "serverVersion", val(iter_item, "serverVersion")
     field "resourceGroup", val(iter_item, "resourceGroup")
-    field "subscriptionId", val(iter_item, "subscriptionId")
```
taterdog590, line 518: for consistency in other policies we have field "id", jmes_path(col_item, "subscriptionId"). Do we want to change this to ID?

XOmniverse: It's not consistent from policy to policy already and it ultimately doesn't actually matter IMO

taterdog590 (approved):

PR #2187, opened 2024-05-08
FOPTS-3939 Update: List of cheaper regions for Azure and AWS vendors

juanestrahcl:
### Description

Issue reported in this support question: https://flexera.atlassian.net/browse/SQ-7976 by Albertsons.
Policy is returning empty results, checking the code it's because some regions for Azure and AWS are outdated, previously we had for example: `US West`, `US West 2`, etc. But now those are named `West US`, `West US 2`, etc. Also few `console.log` statements were removed.

### Issues Resolved

Policy is returning empty results because of outdated regions for Azure and AWS vendors.

https://app.flexera.com/orgs/1105/automation/applied-policies/projects/60073?policyId=663e4547ba767e47966578c9

### Contribution Check List

### DIFF
--- a/cost/flexera/cco/cheaper_regions/README.md
+++ b/cost/flexera/cco/cheaper_regions/README.md
@@ -8,6 +8,8 @@ This Policy Template determines which regions have cheaper alternatives by speci
 
 This Policy Template uses [Credentials](https://docs.flexera.com/flexera/EN/Automation/ManagingCredentialsExternal.htm) for authenticating to datasources -- in order to apply this policy you must have a Credential registered in the system that is compatible with this policy. If there are no Credentials listed when you apply the policy, please contact your Flexera Org Admin and ask them to register a Credential that is compatible with this policy. The information below should be consulted when creating the credential(s).
 
+### Credential Configuration
+
 - [**Flexera Credential**](https://docs.flexera.com/flexera/EN/Automation/ProviderCredentials.htm) (*provider=flexera*) which has the following roles:
   - `billing_center_viewer`
   - `policy_designer`
--- a/cost/flexera/cco/cheaper_regions/README.md
+++ b/cost/flexera/cco/cheaper_regions/README.md
@@ -9,7 +9,7 @@ This Policy Template determines which regions have cheaper alternatives by speci
 This policy has the following input parameters required when launching the policy.
 
 - *Billing Center Name* - The list of Billing centers to check against
-- *Email addresses of the recipients you wish to notify* - A list of email addresses to notify
+- *Email Addresses* - A list of email addresses to notify
 
 ## Policy Actions
 
--- a/cost/flexera/cco/cheaper_regions/README.md
+++ b/cost/flexera/cco/cheaper_regions/README.md
@@ -11,6 +11,12 @@ This policy has the following input parameters required when launching the polic
 - *Billing Center Name* - The list of Billing centers to check against
 - *Email addresses of the recipients you wish to notify* - A list of email addresses to notify
 
+## Policy Actions
+
+The following policy actions are taken on any resources found to be out of compliance.
+
+- Send an email report
+
 ## Prerequisites
 
 This Policy Template uses [Credentials](https://docs.flexera.com/flexera/EN/Automation/ManagingCredentialsExternal.htm) for authenticating to datasources -- in order to apply this policy you must have a Credential registered in the system that is compatible with this policy. If there are no Credentials listed when you apply the policy, please contact your Flexera Org Admin and ask them to register a Credential that is compatible with this policy. The information below should be consulted when creating the credential(s).
--- a/cost/flexera/cco/cheaper_regions/README.md
+++ b/cost/flexera/cco/cheaper_regions/README.md
@@ -1,8 +1,15 @@
 # Cheaper Regions Policy
 
-## What it does
+## What It Does
 
 This Policy Template determines which regions have cheaper alternatives by specifying the expensive region name and the cheaper region name for analysis
+
+## Input Parameters
+
+This policy has the following input parameters required when launching the policy.
+
+- *Billing Center Name* - The list of Billing centers to check against
+- *Email addresses of the recipients you wish to notify* - A list of email addresses to notify
 
 ## Prerequisites
 
@@ -21,13 +28,6 @@ The [Provider-Specific Credentials](https://docs.flexera.com/flexera/EN/Automati
 ## Functional Details
 
 - This policy uses a hash to determine existing regions and newer compatible cheaper regions. It checks the billing center and reports on cheaper regions.
-
-### Input Parameters
-
-This policy has the following input parameters required when launching the policy.
-
-- *Billing Center Name* - The list of Billing centers to check against
-- *Email addresses of the recipients you wish to notify* - A list of email addresses to notify
 
 ## Supported Clouds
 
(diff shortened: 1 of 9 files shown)


XOmniverse (changes requested):
We have a JSON file for AWS regions. We should make a similar file for Azure, and then modify the policy to use those.

/data/aws/regions.json

night-codes (approved):

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -8,30 +8,31 @@ category "Cost"
...
 ###############################################################################
 # Parameters
 ###############################################################################
 
 parameter "param_email" do
   type "list"
+  category "Policy Settings"
   label "Email addresses of the recipients you wish to notify"
+  description "A list of email addresses to notify."
+  # No default value, user input required
```
XOmniverse, line 26: Email parameter should always have a default value of [] unless, for some reason, you expect that nobody would ever use the policy just to see the incident within the Flexera One platform and would *only* want the incident to be emailed.

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -8,30 +8,31 @@ category "Cost"
...
 end
 
 parameter "param_billing_centers" do
-  label "Billing Center Name"
   type "list"
+  category "Filters"
+  label "Billing Center Name"
+  description "List of billing centers to be filtered."
   min_length 1
+  # No default value, user input required
```
XOmniverse, line 35: This should only be a required parameter if Billing Center filtering is required. Is there a reason this policy can't just be run across their entire estate?

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -46,19 +47,29 @@ credentials "auth_flexera" do
 end
 
 ###############################################################################
-# Resources
+# Datasources & Scripts
```
XOmniverse, line 50: The purpose of naming this section "Datasources & Scripts" is that datasources and scripts should be in one section of the policy rather than two sections. Please delete the "Scripts" section and place each script underneath the datasource that is calling it.

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -329,11 +368,11 @@ end
 # Policy
 ###############################################################################
 
-policy "policy_cheaper_regions" do
+policy "pol_cheaper_regions" do
   validate_each $ds_new_bc_cost_name do
     summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{ len data }} rows with cheaper data"
```
XOmniverse, line 373: The pattern we are updating policies to is the following:

`summary_template "{{ with index data 0 }}{{ .policy_name }}{{ end }}: {{ len data }} Cheaper Regions Found"`

policy_name is a field that should be added to the final datasource. You can obtain the policy_name with a ds_applied_policy datasource. Look at the "AWS Unused Volumes" policy if you want an example to use as a reference.

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -329,11 +368,11 @@ end
...
-    check eq(0,1)
-    escalate $email
+    check eq(0, 1)
```
XOmniverse, line 374: The check statement should only raise an incident if something worth reporting was found. A check statement like this will *always* raise an incident, even if no cheaper regions have been found.

XOmniverse (changes requested):

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
juanestrahcl: There is already a logic that, in case that user doesn't enter any `billing_centers` then it will take the top ones (The ones that don't have a parent) so, I removed this comment on code.

cost/flexera/cco/cheaper_regions/README.md
```diff
@@ -1,13 +1,28 @@
...
+## What It Does
 
 This Policy Template determines which regions have cheaper alternatives by specifying the expensive region name and the cheaper region name for analysis
 
+## Input Parameters
+
+This policy has the following input parameters required when launching the policy.
+
+- *Billing Center Name* - The list of Billing centers to check against
+- *Email addresses of the recipients you wish to notify* - A list of email addresses to notify
```
XOmniverse, line 12: This label is too long and doesn't line up with other policies. It should be renamed to just "Email Addresses"

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -8,30 +8,30 @@ category "Cost"
 tenancy "single"
```
XOmniverse, line 8: "tenancy" is an outdated construct and should be removed.

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -322,18 +366,22 @@ script "js_merge_bc_name", type: "javascript" do
...
 ###############################################################################
 
-policy "policy_cheaper_regions" do
-  validate_each $ds_new_bc_cost_name do
-    summary_template "{{ rs_project_name }} (Account ID: {{ rs_project_id }}): {{ len data }} rows with cheaper data"
-    check eq(0,1)
-    escalate $email
+policy "pol_cheaper_regions" do
+  validate $ds_new_bc_cost_name do
+    summary_template "{{ with index data 0 }}{{ .policy_name }}{{ end }}: {{ len data }} Cheaper Regions Found"
```
XOmniverse, line 382: It looks like, after reviewing the incident, that the policy is specifically reporting INSTANCES that can be moved to cheaper regions. The summary_template should always identify specifically the type of resource or item being reported in the incident table. So instead of "Cheaper Regions Found", it should say something like "Instances With Cheaper Region Found".

cost/flexera/cco/cheaper_regions/cheaper_regions.pt
```diff
@@ -244,44 +256,79 @@ script "js_new_costs_request", type: "javascript" do
...
-        "US East":    "US East 2",
-        "KR Central": "KR South",
-        "US Central": "US West 2",
-        "CA Central": "CA East",
-        "EU West":    "UK South",
-        "EU North":   "UK West",
-        "JA East":    "KR South",
-        "JA West":    "KR South"
+    var azure_supported_regions = [
+      { region: "westus", cheaper: "westus2" },
```
XOmniverse, line 269: This kind of data should be stored in the regions.json files for each region. For example, in data/azure/regions.json we can do something like this:

{
    "name": "US West",
    "region": "westus",
    "cheaper": "westus2"
}
  
The general idea is that we should not be hardcoding data like this into policies. Instead, policies should reference one source of truth for this data so that, if it needs to be updated, we only need to update it in one place.

data/azure/regions.json
```diff
@@ -0,0 +1,335 @@
+[
+  {
+    "name": "East US",
+    "region": "eastus",
+    "cheaper": "eastus2"
+  },
+  {
+    "name": "East US 2",
+    "region": "eastus2"
```
XOmniverse, line 9: For the same of consistency, we should include every field in every entry.

For fields that are not relevant, because they don't exist, they should have a null value.

Example:

{
    "name": "South Central US",
    "region": "southcentralus",
    "cheaper": null
  }

XOmniverse (changes requested):

XOmniverse (approved):
LGTM

PR #2194, opened 2024-05-09
POL-1224 Azure RightsizeSQL bug fix

mburmeig:
### Description

Bug was present causing all recommendations to return 0$

### Issues Resolved

Line 743 changed subscriptionId to subscriptionID

https://app.flexera.com/orgs/34088/automation/incidents/projects/135557?incidentId=663b898562db364ad98e80d2

### Contribution Check List

- [x ] New functionality includes testing.
- [x ] New functionality has been documented in CHANGELOG.MD

### DIFF
--- a/cost/azure/rightsize_sql_instances/CHANGELOG.md
+++ b/cost/azure/rightsize_sql_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.2.1
+
+- Fixed a bug causing $0 recommendations
+
 ## v5.2.0
 
 - New fields added to incident output: `Last Accessed` and `Status`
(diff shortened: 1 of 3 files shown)


XOmniverse (changes requested):
When I perused this policy, I noticed a couple other issues (that I am to blame for) related to the naming of this field. I suspect all issues could be fixed by just doing a find/replace and replacing all instances of subscriptionID with subscriptionId

Would you be able to do that, test that the policy still works as expected, and then push that change to this PR?

mburmeig:
I made the changes and tested the policy with success in pepsi.
It was working before and is still working by changing the values of subscriptionID across the entire policy except for where we call subscriptionId from Azure.

XOmniverse (approved):
LGTM

PR #2238, opened 2024-05-20
fix: Meta Parent Consolidated Incidents Only "state=triggered" filter

bryankaraffa:
### Description

Fixes an issue that is causing results from non-current incidents to appear in the Consolidated Incident.. which then reflects an inaccurate resource count.

This fixes an issue with the datasource that gets the incidents for the meta parent policy


XOmniverse (approved):
LGTM

PR #2246, opened 2024-05-22
POL-1244 Update 'resourceType' Incident Fields

XOmniverse:
### Description

This removes the `resourceType` field from policies where this field does not indicate a discrete instance size/capacity/etc. In most cases, it was renamed to the `type` field. This is to prevent issues with how scaped data is presented in the Optimization dashboard.

Additionally, other minor changes were made to bring affected files into compliance with Dangerfile tests.


nia-vf1 (approved):
LGTM

PR #2247, opened 2024-05-23
fix: Meta Parent template `$action_options` check type is array

bryankaraffa:
### Description

Fixes issue where $action_options is null instead of a list, which results in an error


mburmeig (approved):
Looks good

PR #2262, opened 2024-05-28
FOPTS-4009 Revert Flexera Cred Changes in Update README.md

ramyadmz:
### Description

Update Flexera Cred in Read Me file.

### Issues Resolved

We need Optima cred as we call API.

### Contribution Check List

- [ ] New functionality includes testing.
- [ ] New functionality has been documented in the README if applicable
- [ ] New functionality has been documented in CHANGELOG.MD

PR #2263, opened 2024-05-28
Update readme_tests.rb to include colon(`:`) in permission regex.

ramyadmz:
### Description

The current validation for Flexera permission list items in README.md flags correctly formatted items like `common:org:own` (which is the suggested one!) as incorrect. This happens because the existing regex does not allow colons (:).

### Issues Resolved

Updated the regex to include colons `:`

Example of PR that shows this error:
https://github.com/flexera-public/policy_templates/pull/2262

### DIFF
--- a/.dangerfile/readme_tests.rb
+++ b/.dangerfile/readme_tests.rb
@@ -406,7 +406,7 @@ def readme_invalid_credentials?(file)
       fail_message += "```- [**Flexera Credential**](https://docs.flexera.com/flexera/EN/Automation/ProviderCredentials.htm) (*provider=flexera*) which has the following roles:```\n\n"
     end
 
-    flexera_perm_tester = /^`[a-zA-Z0-9\-_\.]+`(?:\*)?$/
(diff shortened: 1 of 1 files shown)


bryankaraffa:
just an fyi if you are seeing unexpected behavior in your #2262 PR related to this you can incorporate any changes required for that to work as expected in that PR... including them together will help validate the issue is actually fixed the change (by including the fix for the script alongside any change to the PT/READMEs you are doing as well)

ramyadmz:
> just an fyi if you are seeing unexpected behavior in your #2262 PR related to this you can incorporate any changes required for that to work as expected in that PR... including them together will help validate the issue is actually fixed the change (by including the fix for the script alongside any change to the PT/READMEs you are doing as well)

Thanks for the suggestion, as this problem looks like a general linter issue, I thought it should be fixed in separate PR
(The PR 2262 is just an example of the problem and any credential with colons in any readme file could face this issue).

PR #2278, opened 2024-05-31
POL-1248 Deprecation Info Field

XOmniverse:
### Description

This adds the following field to the info() block of all deprecated policies: `deprecated: "true"`

Additionally, this adds Dangerfile tests to report errors if this field is missing when the short_description contains the word "deprecated", and when this field is set to true but the short_description does *not* contain the word "deprecated"

Also, a small number of Dangerfile tests will now run even if a policy is deprecated. These tests focus on things that should be checked even for deprecated policies, such as version numbers matching between the policy and the CHANGELOG.

Finally, the active policy automation has been modified to add a "deprecated" field to the active policy list. This field is set to true if the policy is deprecated, and false if it is not.


cost/aws/idle_compute_instances/idle_compute_instances.pt
```diff
@@ -7,11 +7,12 @@ severity "low"
 category "Cost"
 default_frequency "daily"
 info(
-  version: "5.6.1",
+  version: "5.6.2",
```
nia-vf1, line 10: @XOmniverse - does the meta parent policy template version also need to be updated (I know this policy and other policies listed here are deprecated)?

XOmniverse: @nia-vf1 It shouldn't in terms of the meta parents continuing to work, since the policy parameters and incident are unchanged.

nia-vf1 (approved):
LGTM

PR #2359, opened 2024-06-27
POL-1274 Dangerfile: console.log() Test

XOmniverse:
### Description

Adds a new Dangerfile test to raise an error if the console.log() method is invoked. This method is for debugging and should be removed before publishing a policy to the catalog.


nia-vf1 (approved):
LGTM

PR #2360, opened 2024-06-27
POL-1167 p90/95/99 Support for Azure Rightsize Compute/SQL Policies

XOmniverse:
### Description

This adds support for p90/p95/p99 metrics to the `Azure Rightsize Compute Instances` and `Azure Rightsize SQL Databases` policy templates as well as some other changes outlined below:

Azure Rightsize Compute Instances Changelog:
- Added support for p90, p95, and p99 metrics for both CPU and memory.
- Improved calculations for `Minimum` and `Maximum` for both CPU and memory.*

Azure Rightsize SQL Databases Changelog:
- Added `Threshold Statistic` parameter to assess utilization based on various CPU metrics
- Added CPU minimum, maximum, p90, p95, and p99 metrics to incident table

\* Previously, the policy was actually reporting the daily average minimum and maximum CPU usage, since it was averaging the daily metrics returned by the Azure API. It now reports the lowest minimum value and the highest maximum value from those data sets instead, which is more accurate and looks less strange next to the relevant p90/p95/p99 metrics.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=668442998e6ee150dc4b45b8
(`Azure Rightsize SQL Databases` was tested in a client environment where data is available)

### Contribution Check List


cost/azure/rightsize_sql_instances/azure_rightsize_sql_instances.pt
```diff
@@ -592,8 +601,8 @@ script "js_azure_sql_database_metrics", type: "javascript" do
       "api-version": "2018-01-01",
       "timespan": timespan,
       "metricnames": "cpu_percent,connection_successful",
-      "aggregation": "Average,count",
-      "interval":  "P1D"
+      "aggregation": "Average,Minimum,Maximum,count",
+      "interval": "P1D"
```
bryankaraffa, line 605: This should be a parameter imo... The a smaller interval would return more data points, and improve the accuracy of the percentile results.  Default in the Databricks policy template has 15min interval by default

relevant snippets:

The param block:
https://github.com/flexera-public/policy_templates/blob/8adb71e4b2a0b08de688987c159cc9a94f663a80/cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt#L100-L107

The ds to map param value to api payload value:
https://github.com/flexera-public/policy_templates/blob/master/cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt#L384-L405

The request:
https://github.com/flexera-public/policy_templates/blob/master/cost/azure/databricks/rightsize_compute/azure_databricks_rightsize_compute.pt#L587

cost/azure/rightsize_sql_instances/azure_rightsize_sql_instances.pt
```diff
@@ -692,15 +713,51 @@ script "js_merged_metrics", type: "javascript" do
...
+
+        if (cpu_minimum == null) { cpu_minimum = 0 }
 
         if (averages.length > 0) {
           cpu_average = _.reduce(averages, function(memo, num) { return memo + num }, 0) / averages.length
         }
 
-        if (cpu_average <= param_stats_underutil_threshold_cpu_value) {
+        if (minimums.length + maximums.length + averages.length > 0) {
+          cpu_p90 = percentile(minimums.concat(maximums, averages), 90)
```
bryankaraffa, line 742:
```suggestion
          // combine all points from minimums, maximums, averages
          // use percentile() function to get the percentile values from all measured utilization points
          cpu_p90 = percentile(minimums.concat(maximums, averages), 90)
```
Just a comment here to clarify it's already being joined.. my initial thought this was just the `minimums` but then I saw .concat(...)

cost/azure/rightsize_sql_instances/azure_rightsize_sql_instances.pt
bryankaraffa: An example of the results we can get with 1 day interval (only 30 days of data returns * 3 statistics == 90 total points
image

If 9 days of the default 30 lookback period had max:100% cpu, then p90, p90, p99 would == 100.
> .90 * 90 == 81
> For a list of 90 values (min+avg+max measurement * 30 days), the value at index 81 would get selected for p90.. and with 100 being max, so would p95 and p99.
The list of measurements is sorted by maximums so only 9 days out of the 30 would need to have hit max:100% cpu during the day to get this result

With more measurements, we get a better estimate of the underlying distribution. When the sample size is small (e.g., less than 100), percentiles can be sensitive to outliers or extreme values. Larger samples mitigate this issue.

XOmniverse: Would this be within the scope of this ticket, or is this a broader ask that would apply to other policies as well?

Just trying to identify scope and whether this change needs to be part of *this* work or not.

bryankaraffa:
With 15min interval seeing results that make more sense imo:
image

Maybe nit-picking, but perhaps the order of fields in export declaration should be:
Min,Avg,p90,p95,p99,Max (?)

bryankaraffa (approved):

PR #2375, opened 2024-07-01
POL-1278 Deprecate/Unpublish Azure Tag Resources with Resource Group Name Policy

XOmniverse:
### Description

This both deprecates and unpublishes the `Azure Tag Resources with Resource Group Name` policy. This policy does one strange, highly specific thing that doesn't need to be in the catalog; it adds a tag containing the name of the resource group to Azure resources.


nia-vf1 (approved):
LGTM

PR #2616, opened 2024-09-11
POL-1330 New Policy: AWS Account Credentials

XOmniverse:
### Description

This adds a new, unpublished policy template along with a custom meta parent. The purpose of this policy template is to test all of the various cross-account roles implied by an AWS credential to see if they were working as expected or not. Please see the README for more details.

(A custom meta parent is used because the policy engine does not allow you to ignore_status on a signing error, so the meta parent will compare the aggregated incident results to the status of the child policies to determine if the API request succeeded or failed.)

https://app.flexera.com/orgs/27018/automation/applied-policies/projects/119613?policyId=66fab44d7b7d1d58450a795e

(Only one credential in the incident will say "Passed" because equivalent cross-account roles have *not*, in this instance, been created for the other AWS accounts. This result is the expected behavior in that scenario.)

### Contribution Check List


bryankaraffa (commented):
Looks very helpful.  Tested this with a real cred in Customer One and validated we have the IAM Role that exists in each Cloud Account:
https://app.flexera.com/orgs/27018/automation/incidents/projects/119613?incidentId=66e320aa3e9f4ba98626e120

One thought was to potentially create an incident that just contains the accounts that are in error (if any).  Right now. I need to load and filter the results but if there was a separate incident -- if that incident does not exist then
maybe 2 incident summary outcomes, both with the same export table results that contain the full
"`<pt name> - Warning {{ len data }} Cloud Accounts with Credential Errors`"
"`<pt name> - Success 0 Cloud Accounts with Credential Errors`"
detail template could have some additional details if >0 cred errors

Second though, potentially an enhancement for next release as well -- right now using just `sts:GetCallerIdentity` we will find Cloud Accounts where the IAM Role does not exist.  In my experiences if customers use the CloudFormation template and deploy as a StackSet, we usually do not see errors with the IAM Role existing in Cloud Accounts.  Instead more commonly today we see _some Cloud Account(s)_ have regional restrictions and are erroring in one or more regions (but not all).  Maybe we can do a "DescribeInstances" and/or "DescribeVolumes" and/or <<other actions>> on each region, using ignore_status to not fail error -- capturing the regions that don't get back results and returning that to this Policy Incident so we know if a Cloud Account is failing due to a certain region (or all cloud accounts due to the same region which would ellude to SCP)

bryankaraffa:
Getting an error on this from meta parent -- `"Invalid options: no parameter named param_combined_incident_email_child\",`. Debugging now..

bryankaraffa:
Looks like meta parent for this is customized after being compiled.  This would get overwritten by automation and not sure we want to merge this approach?

I presume this is partially (mostly?) because there's >2 incidents for that child PT.. I am looking into modifying meta parent compiler script to better support varying number of incidents in the child PT which will hopefully prevent customization on meta parent afterwords

If there's anything else (like the scenario we saw for Tag Cardinality + Meta Parent) that you are aware of as well which required customization, additional insight would be helpful. Happy to sync over a call as well

bryankaraffa (approved):

PR #2619, opened 2024-09-12
task: remove Databricks from generally recommended

bryankaraffa:
### Description

Databricks has a prerequisite to create Tag Dimension and if a user deploys this PT (i.e. using Onboarding Policy Template) then the PT will error


nia-vf1 (approved):
LGTM

PR #2651, opened 2024-09-20
POL-1362 Flexera Policy CloudFormation Template Automation

XOmniverse:
### Description

This PR enables automatic generation of AWS permissions CFTs in the `tools/cloudformation-template/rolling` directory from permissions listed in the `data/policy_permissions_list/master_policy_permissions_list.json` file. These rolling CFTs include every AWS policy template that is not deprecated, and as a result, the CFT should now include every AWS policy and always be up to date as we add new policies and make changes to existing policies that affect the required permissions.

The "stable release" and release history is not touched. The idea here is that the rolling releases above can be used as a basis for official releases while removing the effort of having to manually maintain those releases in full.

_NOTE: If the user tries to enable individual inline policies for *every* policy template, this exceeds the size limit for AWS. To handle this, a special "All AWS policies" entry is added to the top to create a single inline policy for users that simply want to enable everything. This is still restricted to just the permissions needed for Flexera policies and does not grant read or read/write access to everything across the board._

This also removes a duplicate entry from the validated permissions list and corrects a handful of permissions in README files that were not valid and were caught by the CloudFormation Template linter.

### Testing

The newly generated cloudformation template has been tested in an AWS environment and works as expected.


tools/cloudformation-template/README.md
```diff
@@ -83,7 +89,7 @@ As you follow the official docs, you can use the recommended configurations belo
```
nia-vf1, line 89: I cannot select the line as it is not a part of this change but...

Line 80 says `` - Under **Maximum concurrent accounts**, choose `Percent` and set field value to `100`.``

It should reference 'Failure tolerance' rather than 'Maximum concurrent accounts':
image

Therefore line 80 should read `` -  Under **Failure tolerance**, choose `Percent` and set field value to `100`.``

bryankaraffa (commented):
We should keep revisioning for this PT imo.. for anytime it generates a CFT and there's a diff, create a branch/PR and we can approve it (like the other "automation" workflows we have in this repo).

Also, the publish to S3 workflow logic depends on revisioned copies in the repo:
https://github.com/flexera-public/policy_templates/blob/master/.github/workflows/cfn-publish.yaml
This will need to be changed, or we'll need to keep creating revisioned copied

tools/cloudformation-template/README.md
bryankaraffa: Both Max Concurrent Accounts and Failure tolerance should be set to "Percentage" and "100" for the value

nia-vf1: Yeah but the doc references Max Concurrent Accounts twice, rather than referencing Max Concurrent Accounts once and Failure tolerance once @bryankaraffa

nia-vf1: image
See line 80

tools/cloudformation-template/README.md
```diff
@@ -4,12 +4,16 @@
...
 
+Additionally, two automatically generated rolling release versions are provided but are **not recommended** or supported for production use. These will be used as the basis for the stable releases above.
+
+- [rolling/FlexeraAutomationPolicies.template](https://github.com/flexera-public/policy_templates/blob/master/tools/cloudformation-template/rolling/FlexeraAutomationPolicies.template): Template to add either read or read/action permissions for either all Flexera automation templates or per-Flexera automation template.
+- [rolling/FlexeraAutomationPoliciesReadOnly.template](https://github.com/flexera-public/policy_templates/blob/master/tools/cloudformation-template/rolling/FlexeraAutomationPoliciesReadOnly.template): Identical to the above but with only read only permissions. Recommended when there are concerns over the template having options for more than just read-only access.
```
nia-vf1, line 15:
```suggestion
- [rolling/FlexeraAutomationPoliciesReadOnly.template](https://github.com/flexera-public/policy_templates/blob/master/tools/cloudformation-template/rolling/FlexeraAutomationPoliciesReadOnly.template): Identical to the above but with only read-only permissions. Recommended when there are concerns over the template having options for more than just read-only access.
```

nia-vf1 (approved):
LGTM

PR #2678, opened 2024-09-25
POL-1365 Update Policy Permissions Automation - support for crossed-dagger symbol

nia-vf1:
### Description

For policy permissions in the Readme which have more than two different footnotes, we must use another symbol to differentiate them.

Currently we are using and asterisk and a dagger, but now we have an example of a readme where the permissions have 3 different types of footnotes and therefore this is a change to add the crossed-dagger footnote symbol.

Also, this makes a change to the Dangerfile so that the Pull Request checks also support the new symbol (as well as the existing dagger symbol)

### Contribution Check List

### DIFF
--- a/.dangerfile/readme_tests.rb
+++ b/.dangerfile/readme_tests.rb
@@ -265,7 +265,7 @@ def readme_invalid_credentials?(file, file_lines)
     end
 
     aws_perm_tester = /`[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)*:[a-zA-Z0-9]+`(?:\*)?$/
-    asterix_found = 0
+    footnote_symbol_found = 0
     permission_list_found = 0
 
     aws_permission_text.each_with_index do |line, index|
@@ -277,7 +277,7 @@ def readme_invalid_credentials?(file, file_lines)
         if !line.start_with?("  - ")
           permission_list_found = 2
         else
-          asterix_found = 1 if line.strip.end_with?("*", "†", "‡")
+          footnote_symbol_found = 1 if line.strip.end_with?("*", "†", "‡")
 
           if !line.split("  - ")[1].match?(aws_perm_tester)
             fail_message += "Line #{line_number.to_s}: AWS permission list item formatted incorrectly. Please make sure all list items are formatted like the following examples:\n\n"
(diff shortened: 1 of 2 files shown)


.dangerfile/readme_tests.rb
```diff
@@ -277,31 +279,48 @@ def readme_invalid_credentials?(file, file_lines)
         if !line.start_with?("  - ")
           permission_list_found = 2
         else
-          asterix_found = 1 if line.strip.end_with?("*")
+          footnote_symbols["*"] = true if line.strip.end_with?("*")
+          footnote_symbols["†"] = true if line.strip.end_with?("\u2020")
+          footnote_symbols["‡"] = true if line.strip.end_with?("\u2021")
```
bryankaraffa, line 284: Wondering instead of characters, if we decided to use a different footnote schema.  From wikis, they use
`[^digit]`.

```
This is a sample text with some footnotes [^1][^2][^3].

[1] Some text here
[2] More text
[3] Lorem ipsum
```

With that schema, we can use regex to check/match in the test more easily:
`/\[\^(\d+)\]/`
I think it'll scale out.. we have some PTs with 3 footnotes, but potentially we'll need another in future and how will we decide next character each time?

nia-vf1: I was using this - https://en.wikipedia.org/wiki/Note_(typography)#:~:text=Typographic%20characters%2C%20such%20as%20the,sign%20%C2%A7%2C%20(v)%20the

We can use numbers instead as agree it's probably easier to check (symbols look cooler though😉)

nia-vf1: This will turn into quite a large PR though, as we will need to update all READMEs and also Dangerfile readme script/policy permissions script

bryankaraffa: > I was using this - https://en.wikipedia.org/wiki/Note_(typography)#:~:text=Typographic%20characters%2C%20such%20as%20the,sign%20%C2%A7%2C%20(v)%20the

This works too if we want to minimize scope for this change.. A peer-reviewed list of characters with at least 6 defined checks the box for what we need now imo ✅ 

> (i) the Asterisk [*]
> (ii) the Dagger mark †
> (iii) the Crossed dagger mark ‡
> (iv) the Section sign §
> (v) the Vertical bar ‖
> (vi) the Pilcrow ¶

nia-vf1:
As per this page https://en.wikipedia.org/wiki/Note_(typography)#:~:text=Typographic%20characters%2C%20such%20as%20the,sign%20%C2%A7%2C%20(v)%20the, I have added support for the following footnote symbols:
(i) the Asterisk [*]
(ii) the Dagger mark †
(iii) the Crossed dagger mark ‡
(iv) the Section sign §
(v) the Vertical bar ‖
(vi) the Pilcrow ¶

May move to a number system in future, but this gives us what we need for now

PR #2683, opened 2024-09-26
feat/disable_concurrency_update_active_policy_list_workflow

bryankaraffa:
### Description

Enable only running 1 workflow at a time, cancelling previous in progress

This prevents race-condition between multiple workflows being triggered

We only need one to run at a time, and it should be from the most recent trigger/commit


XOmniverse (approved):
LGTM

PR #2705, opened 2024-10-02
POL-1371 Azure Rightsize NetApp Resources Meta Parent Fix

XOmniverse:
### Description

Fixes an issue causing the consolidated incident in the meta parent to be misnamed.


nia-vf1 (approved):
LGTM

PR #2728, opened 2024-10-08
POL-1379 Azure SQL MI Storage Pricing Automation

XOmniverse:
### Description

This adds automation to gather Azure SQL MI Storage Pricing and updates the Azure SQL MI Storage policy template to use this pricing data. In most cases, this will not matter (the generic SQL DB storage pricing is the same in most cases), but this ensures that any deviations in storage pricing specific to SQL MI are accounted for.


nia-vf1 (approved):
LGTM

PR #2753, opened 2024-10-14
POL-1386 Dangerfile Improvements

XOmniverse:
### Description

* Adds a test for whether verb "GET" is in use
* Updates comma separation test to have fewer false positives


nia-vf1 (approved):
LGTM

PR #2772, opened 2024-10-22
POL-1398 Azure Expiring Credentials - fix Days Until Expiration bug

nia-vf1:
### Description

Regardless of the threshold set, the Azure Expiring Certificates policy returns (in the incident) certificate resources that will expire months and years from now. The reason for this is an incorrect calculation which makes the 'Days Until Expiration' a negative number. This is a change to fix this.

### Issues Resolved

- 'Days Until Expiration' no longer produces a negative value in the policy incident.
- Certificate resources outside of the threshold set are no longer returned in the policy incident.

### Contribution Check List


XOmniverse (approved):
LGTM

PR #2774, opened 2024-10-22
POL-1400 Fix Invalid Recommendations: Azure Rightsize SQL Managed Instances

XOmniverse:
### Description

Azure Rightsize SQL Managed Instances would sometimes produce recommendations for invalid sizes. This is because it was using the existing SQL tier sizes list, and SQL Managed Instances are only available for a much smaller subset of these sizes.

This fixes the issue by creating a separate JSON asset specific to SQL Managed Instances with only the sizes used for that product, and a small modification of the policy template to make use of this new asset.


nia-vf1 (approved):
lgtm

PR #2800, opened 2024-10-31
POL-1408 AWS Account Credentials Meta Not Spawning Children

XOmniverse:
### Description

Fixes issue where the meta policy for AWS Account Credentials was not spawning children due to the policy not properly filtering out meta-specific parameters.


nia-vf1 (approved):
LGTM

PR #2808, opened 2024-11-04
POL-1402 Azure Rightsizing Compute: Consider Managed Disk

XOmniverse:
### Description

This updates the `Azure Rightsize Compute Instances` policy template so that it never recommends a downsize to a size that wouldn't support the current number of attached disks for an instance.

The [Virtual Machine Sizes - List](https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-sizes/list?view=rest-compute-2024-07-01&tabs=HTTP) API call is used during policy execution to get a list of resource types and the maximum number of attached disks. The policy template has been given a major version change since the above API call requires an additional permission.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=6728f54392caab6c320faff1

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #2799, opened 2024-10-30
FOPTS-4819 Added `hide_skip_approvals` to AWS Right Size policies

ramyadmz:
### Description

Adding a new hide_skip_approvals field to the info section of policy templates that do not utilize the request approval option in any escalation. This field enables the UI to dynamically show or hide the "Skip Approval" option based on the policy configuration. Initially, these changes are applied to AWS Right Size policies, with plans to update additional policies in subsequent steps.

More context:[ Seeking Your Input: Proposed Solution for Conditional "Skip Approvals" Visibility](https://teams.microsoft.com/l/message/19:833373548e104af2a20b0216eda1ba7b@thread.skype/1728495063426?tenantId=91034d23-0b63-4943-b138-367d4dfac252&groupId=fb250818-e040-4a26-b207-61c3cd99fd6e&parentMessageId=1728495063426&teamName=Team%20Flexera%20One&channelName=Policy%20Support%20and%20Questions&createdTime=1728495063426)

### Issues Resolved

[FLEX-5397](https://flexera.atlassian.net/browse/FLEX-5397)

### Contribution Check List

- [x] New functionality includes testing.
- [ ] New functionality has been documented in the README if applicable
- [x] New functionality has been documented in CHANGELOG.MD

PR #2813, opened 2024-11-06
FLEX-5397 - Modify Meta Parent Compiler to Conditionally Include `hide_skip_approvals`

ramyadmz:
### Description

This pull request enhances the Meta Parent Policy Template Compiler to conditionally include the hide_skip_approvals field in the generated meta parent policy templates when it exists in the child policy templates.
It also removes prior manual modifications to meta parent policies : https://github.com/flexera-public/policy_templates/pull/2799

More context:[ Seeking Your Input: Proposed Solution for Conditional "Skip Approvals" Visibility](https://teams.microsoft.com/l/message/19:833373548e104af2a20b0216eda1ba7b@thread.skype/1728495063426?tenantId=91034d23-0b63-4943-b138-367d4dfac252&groupId=fb250818-e040-4a26-b207-61c3cd99fd6e&parentMessageId=1728495063426&teamName=Team%20Flexera%20One&channelName=Policy%20Support%20and%20Questions&createdTime=1728495063426)

### Issues Resolved

[FLEX-5397](https://flexera.atlassian.net/browse/FLEX-5397)

### Contribution Check List

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -69,7 +69,8 @@ task :generate_policy_list do
       puts "Skipping #{pp.parsed_name} because publish flag set to a value other than 'true'"
     else
       # Get datetime for last time file was modified
-      updated_at = `git log -1 --format="%cI" -- #{file}`.strip
+      commits = github_client.commits(repo_name, branch, path: file)
+      updated_at = commits.first.commit.author.date.utc.iso8601 if !commits.empty?
       generally_recommended = generally_recommended_template_names.include?(pp.parsed_name) && !deprecated
 
       puts "Adding #{pp.parsed_name}"
--- a/Rakefile
+++ b/Rakefile
@@ -48,12 +48,15 @@ task :generate_policy_list do
       recommendation_type = pp.parsed_info[:recommendation_type]
       publish = pp.parsed_info[:publish]
       deprecated = pp.parsed_info[:deprecated]
+      hide_skip_approvals = pp.parsed_info[:hide_skip_approvals]
+
 
       # 'publish' defaults to true unless explicitly set to false
       # 'deprecated' defaults to false unless explicitly set to true
       publish = !(publish == 'false' || publish == false)
       deprecated = deprecated == 'true' || deprecated == true
-      hide_skip_approvals = pp.parsed_info[:hide_skip_approvals]
+      # 'hide_skip_approvals' defaults to false unless explicitly set to true
+      hide_skip_approvals = hide_skip_approvals == 'true' || hide_skip_approvals == true
     end
 
     # Get version from long description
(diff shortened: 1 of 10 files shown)


Rakefile
```diff
@@ -53,6 +53,7 @@ task :generate_policy_list do
       # 'deprecated' defaults to false unless explicitly set to true
       publish = !(publish == 'false' || publish == false)
       deprecated = deprecated == 'true' || deprecated == true
+      hide_skip_approvals = pp.parsed_info[:hide_skip_approvals]
```
XOmniverse, line 56: Can we confirm that this will work as expected for policy templates that do not have a hide_skip_approvals in their info() block?

XOmniverse (changes requested):

Rakefile
ramyadmz: Good point! Thanks for bringing it up.
I ran the Rakefile locally, and for policy templates without `hide_skip_approvals` in the info() block, the script adds it as nil in the JSON output. This aligns with how other optional fields such as `recommendation_type` are handled, so it should work as expected.
Screenshot 2024-11-08 104542

XOmniverse: For fields that toggle a setting, we want this to reflect what the policy is going to do. In those cases, it should show whatever the default functionality is if the value is not present; in this case, it should be false, similar to what you see with deprecated.

null is only used for non-boolean values that are non-applicable for some policy templates.

ramyadmz: got it, thanks. 
Applied at https://github.com/flexera-public/policy_templates/pull/2813/commits/755c47331075ce29b806cff633e2a1b89696c19d

Rakefile
```diff
@@ -65,8 +69,7 @@ task :generate_policy_list do
       puts "Skipping #{pp.parsed_name} because publish flag set to a value other than 'true'"
     else
       # Get datetime for last time file was modified
-      commits = github_client.commits(repo_name, branch, path: file)
-      updated_at = commits.first.commit.author.date.utc.iso8601 if !commits.empty?
+      updated_at = `git log -1 --format="%cI" -- #{file}`.strip
```
XOmniverse, line 72: What's the reason for changing to a bash command instead of using the built-in github libraries?

XOmniverse (changes requested):

Rakefile
ramyadmz: My bad. I changed it for local testing. revert it: https://github.com/flexera-public/policy_templates/pull/2813/commits/7c7ccccea6e8486660a70d7b13622d116a84cbf4

XOmniverse (approved):
LGTM

PR #2859, opened 2024-11-25
POL-1419 Bug Fix for Azure Publicly-Accessible Blob Containers 

XOmniverse:
### Description

This fixes a bug in the Azure Publicly-Accessible Blob Containers policy template caused by an incorrectly named variable in a JavaScript block.


nia-vf1 (approved):
lgtm

PR #2869, opened 2024-12-05
POL-1414 Currency Conversion Functionality on Kubecost cluster rightsizing policy

joseluiszuflores:
### Description

This update introduces functionality to handle automatic currency conversion within Kubecost cluster rightsizing policy, ensuring that incidents reflect values across different currencies.

#### Changes

* Support for multiple currencies has been added to the system.
* If the kubecost config endpoint returns an empty currency, USD (United States Dollar) will be used as the default currency.
* A condition has been implemented where, if Kubecost's currency matches Flexera's, no currency conversion will take place.

### Issues Resolved

Resolves the issue where currency discrepancies occurred between our template and Kubecost when they had different base currencies.

https://flexera.atlassian.net/browse/POL-1414

Link Applied policy: https://app.flexera.com/orgs/1105/automation/applied-policies/projects/60073?policyId=6751ebfb0f8449d76165faba

### Contribution Check List

### DIFF
--- a/cost/kubecost/cluster/CHANGELOG.md
+++ b/cost/kubecost/cluster/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.3.3
+
+- Added support for multiple currencies: The system now supports handling different currencies.
+- Default currency set to USD: If the kubecost config endpoint returns an empty currency, USD will be used as the default.
+- No conversion when currencies match: If the currency in Kubecost matches the currency in Flexera, no conversion will occur.
+
 ## v0.3.2
 
 - Added `hide_skip_approvals` field to the info section. It dynamically controls "Skip Action Approvals" visibility.
--- a/cost/kubecost/cluster/CHANGELOG.md
+++ b/cost/kubecost/cluster/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.3.3
 
-- Added support for multiple currencies
+- Currency values in incident table are automatically converted to the local currency of the Flexera organization using the most recent exchange rate.
 
 ## v0.3.2
 
--- a/cost/kubecost/cluster/CHANGELOG.md
+++ b/cost/kubecost/cluster/CHANGELOG.md
@@ -2,9 +2,7 @@
 
 ## v0.3.3
 
-- Added support for multiple currencies: The system now supports handling different currencies.
-- Default currency set to USD: If the kubecost config endpoint returns an empty currency, USD will be used as the default.
-- No conversion when currencies match: If the currency in Kubecost matches the currency in Flexera, no conversion will occur.
+- Added support for multiple currencies
 
 ## v0.3.2
 
--- a/cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
+++ b/cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
@@ -108,12 +108,7 @@ datasource "ds_kubecost_currency_code" do
   end
 end
 
-###############################################################################
-# Currencies
-###############################################################################
-
 # Gather local currency info
-
 datasource "ds_currency_target" do
   run_script $js_currency_target, $ds_currency_reference, $ds_currency_code
 end
@@ -172,8 +167,6 @@ datasource "ds_currency_conversion" do
     field "year", jmes_path(response, "year")
   end
 end
-
-#######################
 
 datasource "ds_currency_code" do
   request do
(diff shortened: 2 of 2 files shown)


XOmniverse (changes requested):
Please resolve all warnings/errors returned by Dangerfile on the pull request. If any errors/warnings don't make sense or are false positives, please detail this in the description of the pull request.

cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
```diff
@@ -108,6 +108,73 @@ datasource "ds_kubecost_currency_code" do
   end
 end
 
+###############################################################################
```
XOmniverse, line 111: These style of headers are intended for indicating sections of the policy template, such as "Datasources & Scripts", "Parameters", etc. They shouldn't be used in other ways.

A normal comment should be used here instead.

cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
```diff
@@ -167,7 +239,7 @@ datasource "ds_cluster_sizing" do
   request do
     host $param_kubecost_host
     path "/model/savings/clusterSizing"
-    query "minNodeCount", $param_min_nodes
+    query "minNodeCount", join([$param_min_nodes])
```
XOmniverse, line 242: What is the purpose of the join statement here? Join is used to combine strings, but this only contains one string.

XOmniverse (changes requested):

cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
joseluiszuflores: done!

cost/kubecost/cluster/kubecost_cluster_rightsizing_recommendations.pt
joseluiszuflores: we had inconvinents using only $param_min_nodes:

We initially used minNodeCount, but it seems the request did not include the correct parameter. We're unsure why this occurred.

image


image

then using join statement appears that it's work well

nia-vf1: @joseluiszuflores maybe because the parameter value is a number, not a string? Though I would not expect that to have an impact on it not being able to interpret the query parameter.

joseluiszuflores: looks like our template can't have a number as param let me show  this scenario

I have changed the value by default  to 2 

image

but the same behaviur on template,  we are received empty value

@XOmniverse  @nia-vf1

joseluiszuflores: Note: We updated our template to use the number 2 as a hardcoded value within the request. It seems to be working well.

template: 
image

template applied 
image


We think that  we use numbers like string, at least on request datastore.

example:  
* [name "AWS Savings Plan Recommendations"](https://github.com/flexera-public/policy_templates/blob/b72571abcfc6b13942d1736d30316dcc04fd797a/cost/aws/savings_plan/recommendations/aws_savings_plan_recommendations.pt#L352)
* [AWS Reserved Instances Recommendations](https://github.com/flexera-public/policy_templates/blob/b72571abcfc6b13942d1736d30316dcc04fd797a/cost/aws/reserved_instances/recommendations/aws_reserved_instance_recommendations.pt#L442)
* [Azure Rightsize SQL Managed Instance Storage](https://github.com/flexera-public/policy_templates/blob/b72571abcfc6b13942d1736d30316dcc04fd797a/cost/azure/rightsize_managed_sql_storage/azure_rightsize_managed_sql_storage.pt#L232)
* [Google Rightsize Cloud SQL Recommender](https://github.com/flexera-public/policy_templates/blob/b72571abcfc6b13942d1736d30316dcc04fd797a/cost/google/rightsize_cloudsql_recommendations/google_rightsize_cloudsql_recommendations.pt#L548)

link : Search's [link](https://github.com/search?q=repo%3Aflexera-public%2Fpolicy_templates++query+%22amount%22%2C+%221%22&type=code)

cost/kubecost/cluster/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.3.3
+
+- Added support for multiple currencies
```
XOmniverse, line 5: This should be more specific. I would change it to something like:

Currency values in incident table are automatically converted to the local currency of the Flexera organization using the most recent exchange rate.

XOmniverse (changes requested):

cost/kubecost/cluster/CHANGELOG.md
joseluiszuflores: done!

XOmniverse (approved):
LGTM

PR #2916, opened 2025-01-06
POL-1443 AWS Rightsize Elasticache - Fix undeclared variable causing policy to fail

nia-vf1:
### Description

This change rearranges the code so that the `savings` variable is no longer being referenced before it is declared/defined, meaning the policy no longer fails when it is run.

### Issues Resolved

Fixed an issue with the `savings` variable causing the policy to fail.

Custom policy with these changes have been applied in the customer's tenant and they have confirmed it is now working again as expected.

### Contribution Check List


XOmniverse (approved):
LGTM

PR #2950, opened 2025-02-06
POL-1448 New Policy: FinOps Dashboards

XOmniverse:
### Description

This policy template creates a series of FinOps cloud dashboards within the Flexera organization. Created dashboards are public and can be accessed at the Dashboards -> Cloud page in Flexera One. Optionally, information about the newly created dashboards can be emailed.

Configuration information for these dashboards is stored in separate JSON files located at `data/dashboards/`

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=67c730c06b85ec0517652d24

### Contribution Check List


bryankaraffa (commented):
Instead of 1 file that has all dashboards, should we have each dashboard be it's own JSON file -- with the contents are the exact JSON response you get / put on the dashboards API.
Would make seeing diffs over time easier through git commits, as well as review/deploy those with other methods than the PT.

I also found this repo which appears to be used for a similar purpose... potentially we want to take ownership of that and use that since that data is not 100% PT related.. or we can formally deprecate that repo and update README to point to this new implementation + sources for recommended custom dashboards
https://github.com/flexera-public/cco-dashboards

XOmniverse:
> Instead of 1 file that has all dashboards, should we have each dashboard be it's own JSON file -- with the contents are the exact JSON response you get / put on the dashboards API. May mitigate/prevent issues from modifying the JSON copy+paste, also would enable seeing diffs over time for each dashboard more easily through git commits, as well as review/deploy those with other methods than the PT.
> 
> I also found this repo which appears to be used for a similar purpose... potentially we want to take ownership of that and use that since that data is not 100% PT related.. or we can formally deprecate that repo and update README to point to this new implementation + sources for recommended custom dashboards https://github.com/flexera-public/cco-dashboards

On that second point, my vote would be to simply eliminate that repository. Even this is likely to just be a short-to-medium-term thing as we make changes to move to new dashboarding technology.

On the first point, I could do that, though I question the need for what is, again, likely to just be a short-to-medium-term solution.

nia-vf1 (approved):
LGTM

Can discuss the best place to put the dashboard artifacts in future, for now lets get the policy template added to the repository

PR #3003, opened 2025-03-18
POL-1464 Meta Policy Fix

XOmniverse:
### Description

This is a fix for the meta policy script that prevents the generated meta policy from containing consolidated incidents out of child policy incidents that don't have an export table. 3 malfunctioning meta policies generated by the old version of the script are also being updated using this fixed version of the script.


bryankaraffa:
Confirmed no Applied Policies exist matching name like `Meta Parent:%Scheduled%`.  
👍 looks good to me

PR #3017, opened 2025-03-27
POL-1469 Dangerfile Fixes/Updates

XOmniverse:
### Description

* Updated Ruby and node packages to update versions and solve potential security vulnerabilities.

* Added CodeQL configuration file so that we can enable CodeQL scanning for additional security on PRs.

* Updated Dangerfile "policy_missing_master_permissions?" test to ignore policy templates with `skip_permissions: "true"` in their info block for policy templates with unusual permissions that won't properly scrape or other extenuating reasons we might not want to scape them for permissions.

* Updated Dangerfile "policy_bad_block_name?" test to work when the block name is in single quotes instead of double quotes. Previously, the test would fail to capture a bad block name in this situation.

* Added new Dangerfile "policy_block_name_single_quotes?" test to check if block names are in single quotes instead of double quotes.

* Added new Dangerfile "policy_defunct_metadata?" test to check if defunct metadata fields such as tenancy are used.

* Added new Dangerfile "policy_abbreviated_info_field?" test to ensure consistent naming of info() fields and to avoid abbreviations like CCO, MSP, etc.

* Added new Dangerfile "policy_missing_info_block?" test to report when the info() block is missing entirely. Stolen shamelessly from Bryan.

* Updated existing policy templates so that they will pass the above tests. Also changed these policy templates to pass existing tests as warnings and errors arose from modifying them.

### Contribution Check List


nia-vf1 (approved):
lgtm

PR #3056, opened 2025-04-10
POL-1496 AWS/Azure Instance Types Improvements

XOmniverse:
### Description

This updates the recently-added automation to add AWS and Azure instance types JSON files.

- Instance types are now stored as a flat list. This is because conversation from a list to an object of grouped lists within policy execution is much easier than the reverse.

- Scripts now grab data that can't be pulled from API from the manually maintained instance_types.json files. This eliminates the needs for certain policy templates to use both files and will also make it easier to transition off of them as we find ways to get this data programmatically. 

- "size_rank" field added to AWS. This number corresponds to how many "nanos" of the given instance family a size represents. This is to make it simpler to determine which instance size is above or below another without having to manually parse what nano, medium, 2xlarge, etc. means.


nia-vf1 (approved):
lgtm

PR #3110, opened 2025-04-30
POL-1509 Meta Policy Update: Usability & Incident for Child Policies in Error State

XOmniverse:
### Description

This makes two significant changes:

#### Usability Improvements

- The list of default policy templates to generate meta parents for is now in a separate `default_template_files.yaml` file rather than contained in the script.
- The script now has much improved command line functionality to allow the user to generate meta parents from a custom list or individual policy template files, as well as using a custom meta parent template instead of one of the provided ones if desired.
  - There is also improved error output if the user does not use the script correctly.
- Added improved error handling/output if child policy template is missing necessary modifications to create a meta parent.
- README for the script is updated to include info on the above.

#### Error Incident

This adds a new incident to meta parent policies that list child policies that are in an error state, along with the error details.

The user has the option of deleting these child policies from the incident page so that new ones are generated the next time the meta parent executes. This can eliminate the need to completely terminate/reapply the parent policy if an issue external to Flexera (such as credential permissions) is corrected.

NOTE: The primary thing to review are the updated meta generation templates. The policies themselves are all just generated from those templates.

Tested in an environment with actual data and confirmed to work as expected.

### Contribution Check List


nia-vf1 (approved):
LGTM

PR #3117, opened 2025-05-02
FOPTS-9434 Correctly include all RDS snapshots in policy evaluation

joseluiszuflores:
This PR addresses an issue in the policy template's RDS snapshot filtering logic that was causing some snapshots to be incorrectly excluded from policy evaluation.

### Description

The js_describe_db_snapshots script was incorrectly filtering RDS snapshots. It was grouping snapshots by ParentId (the RDS instance identifier), sorting them by StartTime, and then only retaining the most recent snapshot for each ParentId. This meant that older snapshots, even if they met other policy criteria (like age), were being ignored. This behavior could lead to:

* Data Governance Gaps: Older snapshots, which might be needed for compliance or audit purposes, were not being properly managed by the policy.
* Cost Optimization Inefficiencies: Opportunities to identify and manage older, potentially costly snapshots were being missed.
* Recovery Risks: In some disaster recovery scenarios, the policy might not be able to identify or take action on older snapshots that could be required.

### Issues Resolved

This PR modifies the js_describe_db_snapshots script to correctly include all RDS snapshots in the policy's evaluation. The change involves iterating through all snapshots for each ParentId instead of just selecting the most recent one.

The following code snippet shows the key change in the js_describe_db_snapshots script:

``` golang
--- a/path/to/your/policy/template.rb  # Replace with the actual path
+++ b/path/to/your/policy/template.rb  # Replace with the actual path
@@ -99,10 +99,12 @@
     })
     _.each(_.keys(sorted_by_parent), function(parent) {
       sorted_parents = _.sortBy(sorted_by_parent[parent], "startTime").reverse()
-      result.push(sorted_parents[0])
+      sorted_parents.forEach(function(snapshot) {
+        result.push(snapshot);
+      });
     })
```

* Link SQ
  https://flexera.atlassian.net/browse/SQ-13605
* Link template:
   https://app.flexeratest.com/orgs/1105/automation/projects/60073/policy-templates/681545c3de8432f627dda48c
* Link policy applied:
  https://app.flexera.eu/orgs/34909/automation/applied-policies/projects/136739?policyId=68116186c6ad95b068404016
  
  
[Logs-Applied-policy.log](https://github.com/user-attachments/files/20018972/Logs-Applied-policy.log)

image

### Contribution Check List

### DIFF
--- a/cost/aws/old_snapshots/CHANGELOG.md
+++ b/cost/aws/old_snapshots/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.4.4
+- Corrected incorrect filtering of RDS DB snapshots. The policy was previously only considering the most recent snapshot for each DB instance, leading to missed older snapshots. The logic has been updated to include all snapshots.
+
 ## v8.4.3
 
 - Added `hide_skip_approvals` field to the info section. It dynamically controls "Skip Action Approvals" visibility.
--- a/cost/aws/old_snapshots/CHANGELOG.md
+++ b/cost/aws/old_snapshots/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## v8.4.4
+
 - Corrected incorrect filtering of RDS DB snapshots. The policy was previously only considering the most recent snapshot for each DB instance, leading to missed older snapshots. The logic has been updated to include all snapshots.
 
 ## v8.4.3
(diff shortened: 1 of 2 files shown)


XOmniverse (commented):
Have we verified that these are, in fact, distinct snapshots and not just versions of the same snapshot? From what I recall, the functionality you currently see in the policy template was to avoid producing duplicate results by including every version of a single snapshot in the results.

Do we have a user reporting that entire snapshots are missing from the results that now appear with the change you have made?

joseluiszuflores:
@XOmniverse  Yes, it has been verified that the backups come from the same instance. The customer reported that there are snapshots which, although older than 90 days, were not being included in the previous results. With the changes made to the policy template, those backups are now appearing.

image-20250402-150128
This is the screenshot showing the snapshots that were previously missing.

I'm wondering if we have a justification for why this is happening. What comes to mind is how we manage snapshots: each time a new one is taken, the previous one is no longer considered, as it will always be replaced by a more recent one.

XOmniverse (approved):
Approval but we need to keep an eye on it. I'm still not entirely convinced that this won't introduce redundant recommendations for versioned instances of the same snapshot (which is not the same as different snapshots), and I worry due diligence has not been done to ensure the client understands the distinction.

PR #3154, opened 2025-05-16
POL-1524 Currency Conversion Fix

XOmniverse:
### Description

This fixes an issue with the Currency Conversion policy template where execution would fail if the user built off of the adjustment created by this policy template.

The currency conversion adjustment is now always first in the list being sent into the API, ensuring no issues if later adjustment rules build off of it.

Fix tested in client environment where issue was reported.

### Contribution Check List


tusharshahrs (approved):
LGTM
I checked the Readme and it is fine and does not need to be updated since the changelog is updated.

PR #3179, opened 2025-05-21
task: publish Azure Rightsize NetApp PT

bryankaraffa:
### Description

https://github.com/flexera-public/policy_templates/blob/master/data/active_policy_list/generally_recommended_templates.json#L32
Azure Rightsize NetApp Template should be publish as it's included in the "Recommended Templates" list and we often find savings with this PT during PoCs

Alternative approach to this would be removing this PT from the Recommended PT list.  There are tools that assume all PTs in the list are in the Catalog though so we need to make a change one way or another


XOmniverse (commented):
Policy template does not have publish set to false. This field defaults to true, so this change doesn't actually impact whether the policy template is published or not.

Additionally, when I check the catalog, this policy template is in there.

PR #3251, opened 2025-06-16
POL-1546 AWS Rightsize RDS Instances Better Savings Calculation

XOmniverse:
### Description

The `AWS Rightsize RDS Instances` policy template has been updated to make use of the RDS price list to calculate savings. From the updated README:

- For underutilized resources, the `Estimated Monthly Savings` is calculated based on whether or not the [RDS price sheet](https://raw.githubusercontent.com/flexera-public/policy_templates/refs/heads/master/data/aws/aws_rds_pricing.json) contains list prices for the instance types.
  - If it does, the percentage difference between the list prices of the current instance type and the recommended instance type is multiplied by the full actual cost of the resource. This is then subtracted from the current cost of the resource to calculate the savings.
  - If it does not, the full cost of the resource divided by the number of [NFUs (Normal Form Units)](https://docs.aws.amazon.com/whitepapers/latest/cost-optimization-reservation-models/normalization-factor-for-dedicated-ec2-instances.html) for the current resource size, multiplied by the number of NFUs for the recommended resource size, and then subtracted from the current cost of the resource.
  
The update also corrects some issues that would prevent underutilized resources from appearing in the incident.

### Issues Resolved

We had reports that the estimated savings being reported by this policy template were not accurate enough. This should help with that.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=68506e9b12523d5dae455aed
(Also verified in client environment)

### Contribution Check List

### DIFF
--- a/cost/aws/rightsize_rds_instances/aws_rightsize_rds_instances.pt
+++ b/cost/aws/rightsize_rds_instances/aws_rightsize_rds_instances.pt
@@ -1509,32 +1509,42 @@ script "js_rds_underutil_instance_options", type: "javascript" do
 EOS
 end
 
-datasource "ds_rds_underutil_instances_cleaned" do
-  run_script $js_rds_underutil_instances_cleaned, $ds_rds_underutil_instances, $ds_rds_underutil_instance_options
+datasource "ds_rds_underutil_instance_options_table" do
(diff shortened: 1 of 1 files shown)


tusharshahrs (approved):
LGTM

PR #3305, opened 2025-07-09
POL-1558 New Unpublished CBI Policy Templates

XOmniverse:
### Description

Adds two new unpublished policy templates for sending CSV files from AWS S3 or Azure Blob Storage to Flexera CBI.

Unpublished because these policy templates require some extra knowhow to use, since they send in CSV files over multiple executions, similar to the Oracle CBI policy template. They should generally only be used as a fallback, with guidance from Flexera, when the published policy templates that work in a much more straight-forward and user-friendly way are not able to be used due to the amount of data being sent to CBI.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=686fb68486eaa862688b28ff

### Contribution Check List

### DIFF
--- a/cost/flexera/cco/cbi_ingestion_aws_s3_iterating/CHANGELOG.md
+++ b/cost/flexera/cco/cbi_ingestion_aws_s3_iterating/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## v0.1.0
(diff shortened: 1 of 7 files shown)


tusharshahrs (approved):
LGTM

PR #3332, opened 2025-07-17
SQ-16042 Accommodate missing price for superseded instances.

jc1203:
### Description

Added user-friendly text if the list price for certain instances are not found.

Unfortunately, the list price for a few superseded instances are not return by Azure API, and thus the policy is unable to get such information.

### Issues Resolved

https://flexera.atlassian.net/browse/SQ-16042

### Contribution Check List

### DIFF
--- a/cost/azure/superseded_instances/CHANGELOG.md
+++ b/cost/azure/superseded_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.1.6
+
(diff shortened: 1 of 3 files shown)


XOmniverse (commented):
Is this a change we want to make? There are good reasons to not place text in a field intended for numbers. In particular, it makes manipulating the data in an external tool, like Excel, more difficult, since otherwise you won't be able to do proper sorting and calculations with it.

It would be better to add a disclaimer to the incident description than to add text to a number field.

PR #3347, opened 2025-07-22
POL-1565 Meta Parents: Hourly Schedule

XOmniverse:
### Description

This alters the meta parent templates to default to hourly instead of 15 minutes. This is to help reduce the load on the policy engine and Flexera APIs from people using these templates with their default settings. The user can still select 15 minutes intentionally while applying the meta parents.

Note: The policy templates themselves should update automatically once this change is merged.


tusharshahrs (approved):
LGTM

bryankaraffa (approved):

PR #3365, opened 2025-07-29
POL-1570 Meta Parent: Added Hourly/15 Minute Options

XOmniverse:
### Description

Added "hourly" and "15 minutes" options to meta parent policies. This is primarily for policies like the scheduled instance policies where these options are needed.


taterdog590 (approved):
LGTM thanks!

PR #3449, opened 2025-08-20
POL-1587 Fix Active Policy JSON Automation

XOmniverse:
### Description

Fixes issue with Rakefile that would cause the Active Policy JSON to sometimes have the incorrect "updated_at" datetime for some policy templates. This in turn was affecting the automation to update the policy catalog.


nia-vf1 (approved):
lgtm

PR #3474, opened 2025-08-29
fix: js_make_terminate_request gracefully handle if policy_id not def…

bryankaraffa:
### Description

Updates the `js_make_terminate_request` script to gracefully handle when policy_id is not set (i.e. during retrieve_data).  This snippet is used for "Meta" capabilities in the child policy template, and today it causes an error during local development.

This change, combined with https://github.com/flexera-public/policy_sdk/pull/43 should enable most catalog policy templates to run `retrieve_data` without any need to modify the template, which should improve the policy developers QoL.

https://app.flexera.com/orgs/32682/automation/applied-policies/projects/133896?policyId=68b1ef8045b50292160abfd8


XOmniverse (changes requested):
If we're modifying the policy code (even if it's just a small tweak), we should be changing the version number and updating the CHANGELOG.

I've had good success using AI to generate Python scripts to do this for me. Saves a lot of time.

XOmniverse (approved):
LGTM

PR #3519, opened 2025-09-08
POL-1602 Adding Images to READMEs

XOmniverse:
### Description

Adds more images to various READMEs to show what an incident looks like, focused primarily on policy templates that use image charts.

No testing (apart from making sure the markdown looks correct) since no changes are made to actual code.


nia-vf1 (approved):
lgtm

PR #3607, opened 2025-10-07
POL-1651 Remove "15 minutes" and "Hourly" child schedule options

XOmniverse:
### Description

Removes the "15 minutes" and "Hourly" child schedule options from most meta policies. Adds logic in the meta parent compiler to allow for exceptions if the info() block contains enable_child_schedule_options: "true"


tusharshahrs (approved):
LGTM

PR #3641, opened 2025-10-16
POL-1664 AWS Rightsize EC2 Instances - Capture Missing Memory Stats for Windows Instances in Incident

nia-vf1:
### Description

This change fixes a gap where memory usage data from CloudWatch wasn’t being captured in the policy for many EC2 instances. 

The previous fix seen in https://github.com/flexera-public/policy_templates/pull/3613 fixed the issue for Linux instances, however the problem persisted for Windows instances. This change addresses that.

This fix ensures users now get a more complete picture when reviewing rightsizing recommendations across both CPU and Memory metrics for both Windows and Linux instances.

Tested this in a customer org. Some details from the associated Jira ticket:

> Adding the additional code to support Memory % Used in Bytes metric name and using the new ListMetrics response to build the list of dimensions for GetMetricData fixes this issue.
> 
> Before:
> 
> As we can see, only mem stats for Linux instances are populated.
>
>
> After:
> 
> Now mem stats are populated for both Windows and Linux instances.

### Contribution Check List

### DIFF
--- a/cost/aws/rightsize_ec2_instances/CHANGELOG.md
+++ b/cost/aws/rightsize_ec2_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.4.7
+
+- Resolved issue where memory statistics for Windows instances specifically were not being captured.
+
 ## v5.4.6
 
 - Resolved issue where memory statistics were not being captured.


XOmniverse (approved):
LGTM

PR #3738, opened 2025-11-13
POL-1668 Azure Long Stopped Instances - Fix for currency adjustments 

nia-vf1:
### Description

When observing an applied policy for Azure Long Stopped Instances in a customer org, the policy produces savings amounts in DKK, where the expected behaviour is that it should show the savings amounts in EUR (the currency configuration for their Flexera org).

This is a change to implement the fix for this. The savings amount now takes into account adjustments, such as currency conversion, while ignoring any shared cost related adjustments.

A policy with the change reflected in this PR was applied in the affected customer org and now displays savings amount as expected.

### Contribution Check List

### DIFF
--- a/cost/azure/long_stopped_instances/CHANGELOG.md
+++ b/cost/azure/long_stopped_instances/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v6.1.1
+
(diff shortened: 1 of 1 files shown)


XOmniverse (approved):
LGTM

PR #4317, opened 2026-04-16
POL-1753 "message" Fix

XOmniverse:
### Description

This fixes an issue introduced when bulk modifying policy templates to add CSV support. Inadvertently, "message" and "total_savings" were added to the export block, which isn't logical since this will cause policy execution to fail due to neither of them being in the export block.

This also fixes a small issue with a Dangerfile test, and adds some necessary entries to hash_exclude for a few policy templates that were missing them.


tusharshahrs (approved):
lgtm

PR #4542, opened 2026-06-09
POL-1781 Fix Tag Filtering Logic

XOmniverse:
### Description

Fixed bug in many policy templates where the `!~` exclusion tag operator incorrectly excluded resources whose tag value matched the regex instead of excluding those that did not match

(Dangerfile warnings/errors not related to the above change)


tusharshahrs (approved):
lgtm

PR #4579, opened 2026-06-17
SQ-25756 Google README Fixes

XOmniverse:
### Description

Fixes several Google READMEs that incorrectly list .aggregatedList permissions, which do not exist. These have been replaced with the correct .list permissions instead.

Also updates the Policy API script to handle this correctly going forward.


nia-vf1 (approved):
LGTM

PR #4675, opened 2026-07-07
POL-1795 New Policy Template: Flexera Create Service Account

XOmniverse:
### Description

`Flexera Create Service Account`

This policy template creates a Flexera service account via the IAM API, assigns the specified org-level roles to it, generates a client secret, and then registers an OAuth2 credential in Flexera Automation using that client ID and secret. If all steps succeed, an incident is raised containing the credential details and confirmation of role assignments.

Template is unpublished because it is more intended for internal Flexera use. A Flexeran can add their own token to a new org as a credential, run this to quickly create a service account, and then delete their credential from the Flexera Org, leaving behind a functioning service account that does not require that the client maintain a token associated with someone at their organization to execute policy templates.

(Ignore the dangerfile warning. This is a special case where readability requires something off-spec)

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=6a4cfd41620578afd278f826

### Contribution Check List

### DIFF
--- a/automation/flexera/create_service_account/CHANGELOG.md
+++ b/automation/flexera/create_service_account/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## v0.1.0
+
+- Initial release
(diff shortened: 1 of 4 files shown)


tusharshahrs (approved):
LGTM

PR #4718, opened 2026-07-23
AWS Resources Under or Approaching Extended Support Updates

XOmniverse:
### Description

`AWS Resources Under or Approaching Extended Support`
- A small number of corrections to the JSON file tracking extended support versions.
- Commentary added to the README to inform the user of some caveats:
  - This policy only tracks major versions because it is for tracking costs and should not be used for maintenance/security purposes.
  - The dates are a best effort amalgamation of various sources because AWS does not provide a single consistent source of truth for these dates and often changes them ad hoc via blog posts without updating their alleged authoritative sources of truth.


tusharshahrs (approved):
lgtm

PR #4741, opened 2026-07-29
POL-1807 Meta Parent CWF Logging Fix

XOmniverse:
### Description

Fixes logging issue in meta parent policies where a POST request was incorrectly logged as DELETE. Actual functionality itself works correctly.

(Also used Copilot to do a quick pass for similar logging issues in CWF for non-meta parents but nothing was found)


nia-vf1 (approved):
LGTM

PR #4764, opened 2026-08-05
POL-1811 Flexera Test Support

XOmniverse:
### Description

Adds support for flexeratest.com API endpoints and domains to all relevant policy templates.

Also makes some very minor fixes to a handful of modified policies and their READMEs to fix Dangerfile issues, and makes some improvements to Dangerfile and the Policy API script to avoid false positives.

All remaining Dangerfile errors/warnings are false positives that should be ignored.


tusharshahrs (approved):
lgtm

PR #4767, opened 2026-08-07
POL-1812 Region Check Fix

XOmniverse:
### Description

Fixes issue in several AWS policy templates where the region check would cause policy execution to fail if a 400 response is returned by AWS.


nia-vf1 (approved):
LGTM

PR #4777, opened 2026-08-10
POL-1818 Active Policy JSON: Fix Datetimes

XOmniverse:
### Description

Fixes the updated_at field in the Active Policy JSON to correctly reflect when the PT file was merged into the master branch instead reflecting the final commit.

There were edge cases where technically the final commit for one branch was later than the other, despite being merged earlier, causing the catalog to not update as expected.


tusharshahrs (approved):
lgtm

PR #4812, opened 2026-08-19
POL-1826 Parameter Sanitization: Tooling

XOmniverse:
### Description

The first of several PRs focused on sanitizing inputs for PT parameters. This is to prevent users from breaking policy template execution because they entered a parameter incorrectly, such as putting whitespace at the beginning or end of the value.

This PR updates tooling around our PTs, such as scripts, meta parent generation, and the policy dev agent itself.


tusharshahrs (approved):
lgtm

PR #4871, opened 2026-08-26
POL-1828 AWS S3 Buckets Without Lifecycle Configuration: Bug Fix

XOmniverse:
### Description

AWS S3 Buckets Without Lifecycle Configuration - Fixed bug where the policy would fail with a `'resource' is not defined` error whenever the `Exclusion Tags` parameter was used.

Also fixed bugs in this and two other policy templates related to the tag filtering not working correctly.

https://app.flexera.com/orgs/6/automation/applied-policies/projects/7954?policyId=6a8f3d1e30ae6a2d17bf6332

### Contribution Check List

### DIFF
--- a/cost/aws/s3_lifecycle/aws_s3_lifecycle.pt
+++ b/cost/aws/s3_lifecycle/aws_s3_lifecycle.pt
@@ -490,7 +490,7 @@ script "js_aws_buckets_tag_filtered", type: "javascript" do
   })
 
   if (param_exclusion_tags.length > 0) {
-    result = _.filter(aws_buckets, function(bucket) {
+    result = _.reject(aws_buckets, function(bucket) {
       resource_tags = {}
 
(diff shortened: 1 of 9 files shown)


cost/aws/s3_lifecycle/aws_s3_lifecycle.pt
```diff
@@ -493,8 +493,8 @@ script "js_aws_buckets_tag_filtered", type: "javascript" do
     result = _.filter(aws_buckets, function(bucket) {
```
nia-vf1: Technically unrelated to this fix but should this be `_.reject` as opposed to `_.filter` ?

For example:

```suggestion
    result = _.reject(aws_buckets, function(bucket) {
```

Can look into this separately as the current change addresses the bug fix.

nia-vf1 (approved):
LGTM

PR #4974, opened 2026-09-11
POL-1841 Content-Type Update

XOmniverse:
### Description

Updates all policy templates to use "Content-Type" instead of "content-type" in API request headers. This is to ensure that the policy engine properly masks the default type of "text/plain" when this header is specified explicitly in the policy template; this masking is currently case sensitive.


tusharshahrs (approved):
LGTM

PR #4982, opened 2026-09-11
POL-1843 Content-Type Dangerfile Test

XOmniverse:
### Description

Add a Dangerfile test to make sure Content-Type headers always have the correct casing.

Tested and verified that the test works as expected.


tusharshahrs (approved):
lgtm

PR #5017, opened 2026-09-17
FOPTS-29760 Migrated to Azure Resource Graph API for retrieving Azure Storage Accounts.

jc1203:
### Description

Migrated to Azure Resource Graph API for retrieving Azure Storage Accounts.

This should eliminate the 429 throttling issue.

### Jira tickets

https://flexera.atlassian.net/browse/FOPTS-29760
https://flexera.atlassian.net/browse/SQ-27993

**TODO**

### Contribution Check List


XOmniverse (changes requested):
Mostly looks good. Some feedback:

1) This should be a MAJOR version change, not MINOR, because it is a breaking change. Users will need to modify credentials to keep using these policy templates as-is. See: https://github.com/flexera-public/policy_templates/blob/master/STYLE_GUIDE.md#versioning

2) Numerous Dangerfile warnings were produced due to changing permissions. Make sure to double check that the READMEs correctly list the permissions that are actually needed after making these changes. If you've done so and the warnings are false positives, note that in the PR description.

nia-vf1:
Would definitely like to test this one across a few customer orgs before approving this.

jc1203:
Hi @XOmniverse, @nia-vf1 

I have just bumped all the major versions as you have pointed out.
Note that there were 3 policies with pre-release version like "0.x.y", and I bumped them to "1.0.0".
(deprecated_storage_accounts 0.4.3, data_lake_optimization 0.4.4, unused_storage_accounts 0.5.2)

Double checked the permissions, and I believe that the Danger file is false-positive. Added note in PR regarding Danger checks.

Tested the policies on both staging as well as customer environments, please see links to the applied policies in PR description.

PR #5021, opened 2026-09-18
POL-1846 Meta Parent Blank Value Bug Fix

XOmniverse:
### Description

Fixes issue where meta parents would raise children with empty values for cloud account if the Optima API returns any costs without a value for vendor_account.


tusharshahrs (approved):
lgtm

